Issue #16, opened 2018-01-05
Dependencies with licensing problems


jinwoo:
I'm adding license checking to Google Node team's repos using our [`js-green-licenses`](https://github.com/google/js-green-licenses) tool, and finds out that nodejs-common has two problematic dependencies: log-driver & modelo.

[log-driver](https://www.npmjs.com/package/log-driver) doesn't specify any licenses. A 3-year-old [commit](https://github.com/cainus/logdriver/commit/7caec881ea7e1082c6733218398ad45650fa850b) added the ISC license and bumped its version to 1.2.6 but no NPM packages have been published with that change.

[modelo](https://www.npmjs.com/package/modelo) is in a better situation. It has the LICENSE file in its repo but its `package.json` doesn't specify its license and js-green-licenses reports that as an error.

I filed an issue for log-driver: https://github.com/cainus/logdriver/issues/9, and a PR for modelo: https://github.com/kevinconway/Modelo.js/pull/20

This issue is for tracking those problems. Once they are resolved and new npm packages are published, we have to adjust the version numbers for them accordingly.

Until then, all our packages that depend on nodejs-common must have a package whitelist for `js-green-licenses`.

jinwoo:
kevinconway/Modelo.js#20 has been merged and a new version of npm ([v4.2.3](https://www.npmjs.com/package/modelo)) has been published.

No responses to cainus/logdriver#9 yet though.

stephenplusplus:
log-driver@1.2.6 seems to have been published, however, jsgl isn't happy still:

```
> @google-cloud/common@0.16.0 posttest C:\Users\sawch\dev\nodejs-common
> npm run license-check


> @google-cloud/common@0.16.0 license-check C:\Users\sawch\dev\nodejs-common
> jsgl --local .

Checking package.json...

(no license): forever-agent@0.5.2
  @google-cloud/common@0.16.0 -> log-driver@1.2.6 -> codecov.io@0.0.1 -> request@2.42.0 -> forever-agent@0.5.2

(no license): combined-stream@0.0.7
  @google-cloud/common@0.16.0 -> log-driver@1.2.6 -> codecov.io@0.0.1 -> request@2.42.0 -> form-data@0.1.4 -> combined-stream@0.0.7

(no license): delayed-stream@0.0.5
[...]

PR #18, opened 2018-01-08
chore: license check in posttest

jinwoo:
The package, `log-driver`, should be whitelisted. See
https://github.com/googleapis/nodejs-common/issues/16.

### DIFF
--- a/js-green-licenses.json
+++ b/js-green-licenses.json
@@ -1,0 +1,5 @@
+{
+  "packageWhitelist": [
+    "log-driver"  // https://github.com/googleapis/nodejs-common/issues/16
+  ]
+}
--- a/package.json
+++ b/package.json
@@ -40,7 +40,9 @@
     "cover": "nyc --reporter=lcov mocha --require intelli-espower-loader test/*.js && nyc report",
     "samples-test": "cd samples/ && npm link ../ && npm test && cd ../",
     "test-no-cover": "repo-tools test run --cmd mocha -- test/*.js --no-timeouts",
-    "test": "repo-tools test run --cmd npm -- run cover"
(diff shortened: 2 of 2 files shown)


js-green-licenses.json
```diff
@@ -0,0 +1,5 @@
+{
+  "packageWhitelist": [
+    "log-driver"  // https://github.com/googleapis/nodejs-common/issues/16
```
stephenplusplus: The comment makes the JSON invalid :(

package.json
```diff
@@ -40,7 +40,9 @@
     "cover": "nyc --reporter=lcov mocha --require intelli-espower-loader test/*.js && nyc report",
     "samples-test": "cd samples/ && npm link ../ && npm test && cd ../",
     "test-no-cover": "repo-tools test run --cmd mocha -- test/*.js --no-timeouts",
-    "test": "repo-tools test run --cmd npm -- run cover"
+    "test": "repo-tools test run --cmd npm -- run cover",
```
stephenplusplus: I wonder if it's possible to put the license check inside of the repo-tools command? @jmdobry ?

js-green-licenses.json
jinwoo: `jsgl` allows comments here, by stripping them out before parsing :)

package.json
jinwoo: I'm fine with making repo-tools support license checking but I think that is beyond this PR's scope :)

Any objections to committing this PR as is?

stephenplusplus: Issue opened to track that request: https://github.com/GoogleCloudPlatform/nodejs-repo-tools/issues/94

jinwoo: Thanks, @stephenplusplus !

PR #25, opened 2018-01-22
Support Stream constructor options in pagination methods.

### LINKED ISSUE #24: Provide a way to set `highWaterMark` in streams
#### Environment details

  - OS: All
  - Node.js version: 8.x
  - npm version: 5.x
  - @google-cloud/common version: 0.15.1

#### Steps to reproduce

  1. Try to set the `highWaterMark` on `bigquery.createQueryStream` or any other stream in the Google Cloud packages.
  2. You can't. When the limiter object is instantiated, [it passes nothing to through2](https://github.com/googleapis/nodejs-common/blob/master/src/util.js#L640). This would be its opportunity to set `highWaterMark`.

Setting `highWaterMark` allows developers to have some control over how much memory is used in memory-sensitive situations.

stephenplusplus:
Fixes #24

This will allow all paginated streaming methods in the various `@google-cloud/*` libraries to customize the stream options, most popularly "highWaterMark" (used to control backpressure).

### DIFF
--- a/src/paginator.js
+++ b/src/paginator.js
@@ -107,7 +107,6 @@ paginator.parseArguments_ = function(args) {
   var autoPaginate = true;
   var maxApiCalls = -1;
   var maxResults = -1;
-  var streamOptions;
   var callback;
 
   var firstArgument = args[0];
@@ -159,7 +158,6 @@ paginator.parseArguments_ = function(args) {
 
   parsedArguments.streamOptions = extend(true, {}, parsedArguments.query);
   delete parsedArguments.streamOptions.autoPaginate;
-  delete parsedArguments.streamOptions.maxApiCalls;
   delete parsedArguments.streamOptions.maxResults;
   delete parsedArguments.streamOptions.pageSize;
 
--- a/src/paginator.js
+++ b/src/paginator.js
@@ -107,6 +107,7 @@ paginator.parseArguments_ = function(args) {
   var autoPaginate = true;
   var maxApiCalls = -1;
   var maxResults = -1;
+  var streamOptions;
   var callback;
 
   var firstArgument = args[0];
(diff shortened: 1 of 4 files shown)


src/paginator.js
```diff
@@ -148,13 +149,21 @@ paginator.parseArguments_ = function(args) {
...
     maxApiCalls: maxApiCalls,
     maxResults: maxResults,
     callback: callback,
   };
+
+  parsedArguments.streamOptions = extend(true, {}, parsedArguments.query);
+  delete parsedArguments.streamOptions.autoPaginate;
+  delete parsedArguments.streamOptions.maxApiCalls;
+  delete parsedArguments.streamOptions.maxResults;
+  delete parsedArguments.streamOptions.pageSize;
```
jimkang: `autoPaginate`, `maxApiCalls`, etc. seem to be properties of `parsedArguments` itself, rather than `parsedArguments.query`. So, if `parsedArguments.streamOptions` comes from copying `parsedArguments.query`, wouldn't those properties be absent from `parsedArguments.streamOptions` and thus not need to be deleted?

Other than that, this LGTM!

src/paginator.js
```diff
@@ -107,6 +107,7 @@ paginator.parseArguments_ = function(args) {
   var autoPaginate = true;
   var maxApiCalls = -1;
   var maxResults = -1;
+  var streamOptions;
```
alexander-fenster: is it used anywhere?

src/paginator.js
stephenplusplus: That's true, however, they originate from `query`. After your comment though, I did find one thing we don't need to delete, "maxApiCalls", as that is deleted in earlier code.

src/paginator.js
stephenplusplus: Nope, and now it's toast.

src/paginator.js
jimkang: Good to know!

jimkang:
Anyone have any idea when this will be published in @google-cloud/bigquery?

PR #32, opened 2018-02-08
Do not allow the project ID token to be sent to google-auto-auth.

### LINKED ISSUE #31: Missing projectId after upgrading to 0.16.0
#### Environment details

  - OS: Linux (running on GCE instance)
  - Node.js version: 8.9.4
  - npm version: 5.6.0
  - @google-cloud/common version: 0.16.0

#### Steps to reproduce
I updated the version of google-cloud/common to 0.16.0 in google-cloud/profiler (repo [here](https://github.com/GoogleCloudPlatform/cloud-profiler-nodejs)). Then when running the profiler on a GCE instance without specifying the projectId, I got this error message:

Error: Sorry, we cannot connect to Cloud Services without a project ID. You may specify one with an environment variable named "GCLOUD_PROJECT". See https://googlecloudplatform.github.io/google-cloud-node/#/docs/guides/authentication for a detailed guide on creating an authenticated connection.

I switched the version back to 0.15.1 of google-cloud/common and no longer saw this error.

I'm not sure if this is an intended change with the latest release, but it felt surprising to me.

stephenplusplus:
Fixes #31 

This will block google-auto-auth from receiving `{{projectId}}`, which it assumes is the user's project ID.

### DIFF
--- a/test/util.js
+++ b/test/util.js
@@ -689,6 +689,19 @@ describe('common/util', function() {
 
       googleAutoAuthOverride = function(config_) {
         assert.strictEqual(config_.projectId, undefined);
+        setImmediate(done);
+        return authClient;
+      };
+
+      util.makeAuthenticatedRequestFactory(config);
+    });
+
+    it('should not remove projectId from config object', function(done) {
+      var config = {
+        projectId: '{{projectId}}',
+      };
+
+      googleAutoAuthOverride = function(config_) {
         assert.strictEqual(config.projectId, '{{projectId}}');
         setImmediate(done);
         return authClient;
--- a/test/util.js
+++ b/test/util.js
@@ -701,7 +701,7 @@ describe('common/util', function() {
         projectId: '{{projectId}}',
       };
 
-      googleAutoAuthOverride = function(config_) {
+      googleAutoAuthOverride = function() {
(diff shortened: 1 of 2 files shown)


test/util.js
```diff
@@ -669,10 +669,27 @@ describe('common/util', function() {
...
+
+    it('should not pass projectId token to google-auto-auth', function(done) {
+      var config = {
+        projectId: '{{projectId}}',
+      };
 
       googleAutoAuthOverride = function(config_) {
-        assert.strictEqual(config_, config);
+        assert.strictEqual(config_.projectId, undefined);
+        assert.strictEqual(config.projectId, '{{projectId}}');
```
alexander-fenster: this check is not needed: we're testing that `{{projectId}}` is not passed down, not that it was originally assigned correctly.

alexander-fenster: or is it here to verify that the caller's object is kept unchanged? if so, it should be in a different small unit test.

stephenplusplus: This is a confirmation that we didn't delete it from the real config object.

alexander-fenster: Yeah, then the name `'should not pass projectId token to google-auto-auth'` is misleading.  (but it's a minor thing of course)

stephenplusplus: Split 'em!

PR #34, opened 2018-02-10
fix: pin log-driver to 1.2.5

alexander-fenster:
The first time this year where greenkeeper CI check proved useful.


alexander-fenster:
Probably it does not like the fact that the PR author and commit author are different persons. Since both @greenkeeper[bot] and myself signed the CLA, should be fine.

PR #48, opened 2018-03-07
chore: run appveyor on node8 only

alexander-fenster:
We don't run it on node4, 6, 7 for other libraries, no reason to do it in this library as well.

- [x] Tests and linter pass

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,8 +1,5 @@
 environment:
   matrix:
-  - nodejs_version: 4
-  - nodejs_version: 6
-  - nodejs_version: 7
   - nodejs_version: 8
 
 install:


callmehiphop:
Interesting, for some reason I thought we did run it on all major versions? Any reason we don't?

alexander-fenster:
@callmehiphop My understanding is that it's just too slow (we only run 1 build in parallel). Even with just node8 it takes much more time than Circle takes for running tests on all 4 supported versions.

alexander-fenster:
Anyway, the correct action would be either make all libraries run windows builds on all versions, or make this library run windows builds only on node8 - I see no reason of having different setup for different packages :)

callmehiphop:
I totally agree! I was just curious about why we only run 8

alexander-fenster:
As I said, the only reason I can think about is that it's too slow to wait for 4 versions :(
Thanks!

PR #57, opened 2018-03-19
Support access token

stephenplusplus:
Fixes #11 

This will allow a user to provide an access token, which will then be used by google-auto-auth.

PR #62, opened 2018-03-21
0.17.0

stephenplusplus:
## Features

- (#57): Accept an access token to a Service instantiation.
- (#63): Privatize all instance properties.

## Fixes

- (#61): When returning an error about a missing project ID, the stacktrace will be preserved.


alexander-fenster:
How about including #63 in this release as well?

PR #64, opened 2018-03-25
Convert to TypeScript

JustinBeckwith:
This PR is split into 2 commits:  the first does the js->ts rename, the second gets compilation going.  This is the first in a number of steps to get us movin with TypeScript.


src/paginator.ts
```diff
@@ -112,13 +110,13 @@ paginator.parseArguments_ = function(args) {
   const firstArgument = args[0];
   const lastArgument = args[args.length - 1];
 
-  if (is.fn(firstArgument)) {
+  if ((is as any).fn(firstArgument)) {
```
stephenplusplus: As a non-TS user, could you break down what's going on here? And if there's a prettier solution, or if we should consider a different library than `is`?

JustinBeckwith: Yeah, this was super ugly.  The `@types/is` package doesn't have `is.fn` for some reason.  This will go away as soon as this lands:
https://github.com/DefinitelyTyped/DefinitelyTyped/pull/24542

stephenplusplus: Ah, cool. Thanks!

JustinBeckwith:
@jmdobry how do we want to handle docs with this?

package.json
```diff
@@ -37,12 +37,13 @@
     "docs": "repo-tools exec -- jsdoc -c .jsdoc.js",
     "lint": "repo-tools lint --cmd eslint -- src/ samples/ system-test/ test/",
     "prettier": "repo-tools exec -- prettier --write src/*.js src/*/*.js samples/*.js samples/*/*.js test/*.js test/*/*.js system-test/*.js system-test/*/*.js",
```
alexander-fenster: do we need `gts check` somewhere in the scripts?

JustinBeckwith: Good question :) My plan was to land this first.  `gts fix` is going to make the changeset *huge*, and I want to isolate that to one PR.

PR #67, opened 2018-03-28
chore: use arrow functions

JustinBeckwith:
Fairly routine find & replace to catch most of the places where we can use arrow functions.  Not done yet, but a good start. Rubber stamp appreciated :)

### DIFF
--- a/src/operation.ts
+++ b/src/operation.ts
@@ -92,8 +92,8 @@ modelo.inherits(Operation, ServiceObject, events.EventEmitter);
 Operation.prototype.promise = function() {
   const self = this;
 
-  return new self.Promise(function(resolve, reject) {
-    self.on('error', reject).on('complete', function(metadata) {
+  return new self.Promise((resolve, reject) => {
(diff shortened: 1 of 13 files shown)


src/operation.ts
```diff
@@ -173,7 +173,7 @@ Operation.prototype.startPolling_ = function() {
     return;
   }
 
-  this.poll_(function(err, metadata) {
+  this.poll_((err, metadata) => {
     if (err) {
       self.emit('error', err);
```
callmehiphop: I'm not sure if this is something you'd want to tackle here as well, but you should be able to remove a lot of the `self` variables in favor of `this` if we're using arrow functions now.

JustinBeckwith: Absolutely!  That's something I was going to address in a follow up PR.  Wanted to keep this one nice and simple :)

callmehiphop:
I know this probably is not the best place to start this conversation, but I think with all the incoming changes now would be a great time to change how we denote private methods - `myMethod_` :arrow_right: `_myMethod`

PR #70, opened 2018-03-31
fix: ensure packaging works

JustinBeckwith:
I broke this with #64

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,9 +8,11 @@
     "node": ">=4.0.0"
   },
   "repository": "googleapis/nodejs-common",
-  "main": "./src/index.js",
+  "main": "./build/src/index.js",
+  "types": "./build/src/index.d.ts",
   "files": [
-    "src",
+    "build/src",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -8,9 +8,11 @@
     "node": ">=4.0.0"
   },
   "repository": "googleapis/nodejs-common",
-  "main": "./src/index.js",
+  "main": "./build/src/index.js",
+  "types": "./build/src/index.d.ts",
   "files": [
-    "src",
+    "build/src",
+    "package.json",
```
jinwoo: `package.json` is automatically included, but explicitly listing it looks good to me.

The [document](https://docs.npmjs.com/files/package.json#files) on npmjs.com says `README` is also always included but I'm not sure if it applies to `README.md`. Can you check or list it here too?

stephenplusplus: It does. READMEs, LICENSEs, and package.json are auto-included 👍

PR #71, opened 2018-03-31
chore: types, types, types

JustinBeckwith:
Not complete at all, but a good start towards typing

### DIFF
--- a/src/logger.ts
+++ b/src/logger.ts
@@ -52,25 +52,27 @@ interface LoggerOptions {
  * @param {Array.<string>=} options.levels - The list of levels to use. (Default: logger.LEVELS)
  * @param {string=} options.tag - A tag to use in log messages.
  */
-function logger(options?: LoggerOptions|string) {
-  let opts: LoggerOptions;
-  if (is.string(options)) {
-    opts = {
-      level: options as string,
+function logger(options?: LoggerOptions);
+function logger(level?: string);
+function logger(optionsOrLevel?: LoggerOptions|string) {
+  let options: LoggerOptions;
+  if (is.string(optionsOrLevel)) {
+    options = {
+      level: optionsOrLevel as string,
     };
   } else {
-    opts = (options || {}) as LoggerOptions;
+    options = (optionsOrLevel || {}) as LoggerOptions;
   }
 
   return logDriver({
-    levels: opts.levels || LEVELS,
+    levels: options.levels || LEVELS,
 
-    level: opts.level || 'error',
+    level: options.level || 'error',
 
     format() {
       const args = [].slice.call(arguments);
       const level = args.shift().toUpperCase();
-      const tag = opts.tag ? ':' + opts.tag + ':' : '';
+      const tag = options.tag ? ':' + options.tag + ':' : '';
       const message = args.join(' ');
       return `${level}${tag} ${message}`;
     },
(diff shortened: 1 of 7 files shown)


JustinBeckwith:
hah, oh yeah - I need to wait until https://github.com/DefinitelyTyped/DefinitelyTyped/pull/24650 lands :)

JustinBeckwith:
This is ready for a looksy now 👍

JustinBeckwith:
Sweet.  @callmehiphop @alexander-fenster mind taking a gander?

JustinBeckwith:
@googleapis/node-team PTAL :)

src/operation.ts
```diff
@@ -26,7 +26,21 @@ const modelo = require('modelo');
  * @type {module:common/serviceObject}
  * @private
  */
-const ServiceObject = require('./service-object.js');
+const ServiceObject = require('./service-object');
+
+export interface OperationConfig {
```
kjin: Can you add a TODO to make these fields more descriptive/possibly optional?

src/logger.ts
```diff
@@ -27,35 +27,50 @@ const logDriver = require('log-driver');
...
- * @param {Array.<string>=} options.levels - The list of levels to use. (Default:
- *     logger.LEVELS)
+ * @param {string=|object=} options - Configuration object. If a string, it is treated as `options.level`.
+ * @param {string=} options.level - The minimum log level that will print to the console. (Default: `error`)
+ * @param {Array.<string>=} options.levels - The list of levels to use. (Default: logger.LEVELS)
  * @param {string=} options.tag - A tag to use in log messages.
  */
-function logger(options) {
+function logger(options?: LoggerOptions|string) {
+  let opts: LoggerOptions;
```
kjin: Does
```ts
if (is.string(options)) {
  options = { level: options as string };
}
```
not work? We should prefer not to have `options` and `opts` in the same scope.

src/operation.ts
```diff
@@ -141,8 +155,8 @@ Operation.prototype.listenForEvents_ = function() {
  *
  * @param {function} callback
  */
-Operation.prototype.poll_ = function(callback) {
-  this.getMetadata((err, resp) => {
+Operation.prototype.poll_ = function(callback: (err?: Error|null, resp?: any) => void) {
```
kjin: Nit: We should standardize the callback type pattern here and elsewhere. I recommend `(err: Error|null, resp?: any) => void` (though parameterizing the function on the type of resp would be better).

src/service-object.ts
```diff
@@ -149,52 +197,60 @@ ServiceObject.prototype.delete = function(callback) {
 
   // The `request` method may have been overridden to hold any special behavior.
   // Ensure we call the original `request` method.
-  ServiceObject.prototype.request.call(this, reqOpts, (err, resp) => {
+  ServiceObject.prototype.request.call(this, reqOpts, (err: Error|null, resp: r.Response) => {
```
kjin: Same comment as above regarding callbacks.

src/logger.ts
JustinBeckwith: Fixed this with proper TypeScript method overloading.  Good call.

src/operation.ts
JustinBeckwith: I hear you.  Step 1 is to kind of identify what the various return types are across the module.  It's a little scattered and unclear.  After we're confident we have all the right types in place, I'm 100% for distilling it down to a rational set of types.

Issue #13, opened 2017-12-13
Hide extraneous GCN plumbing from returned objects


stephenplusplus:
<table><th colspan=2>Copied from <a href="https://github.com/GoogleCloudPlatform/google-cloud-node/issues/2217">GoogleCloudPlatform/google-cloud-node#2217</a></th><tr><td width=70><a href="/adam-sandor"><img src="https://avatars2.githubusercontent.com/u/10675791?s=88&v=4" height=44 width=44></a></td><td name="20170413114442000Z">@adam-sandor<br><a href="#20170413114442000Z">April 13, 2017 11:44</a></td></tr><tr><td colspan=2>

Fetching a list of files from my bucket produces 271K of JSON output from the server. And I just have 4 files in the bucket.
The data includes things like the complete description of the bucket itself (per every file) or the description of the googlecloud storage library (authors, dependencies, etc). Some of this is probably not loaded from the server but is added by the node api library. Even if that doesn't lead to network traffic it's still unnecessary and inconvenient when I try to find the relevant data. The screenshot below show the data structure with most elements collapsed. None of those elements are relevant for listing files in the bucket. I only want to see the data that is printed by `gsutil` when doing `ls`.

[https://www.dropbox.com/s/4dgpu7jh8z6lb9h/Screen%20Shot%202017-04-13%20at%2013.29.58.png?dl=0](https://www.dropbox.com/s/4dgpu7jh8z6lb9h/Screen%20Shot%202017-04-13%20at%2013.29.58.png?dl=0)

I might be missing some option or parameter that would limit the output from getFiles. 

#### Environment details

  - OS: OSX
  - Node.js version: 7.2.0
  - npm version: 3.10.9
  - google-cloud-node version: 1.0.0

#### Steps to reproduce

```
var storage = require('@google-cloud/storage');
var fs = require('fs');

var gcs = storage({
    projectId: 'MYPROJECT'
});

var bucket = gcs.bucket('MYBUCKET');

bucket.getFiles(function(err, files) {
    if (!err) {
        fs.writeFileSync("files.json", JSON.stringify(files))
    } else {
        console.error(err)
    }
});
```

[...]

Issue #53, opened 2018-03-14
Switch to google-auth-library


JustinBeckwith:
With some of the recent additions to google-auth-library. I think we can switch this lib off of google-auto-auth.  It may be nice to eliminate the middle-lib here. 

@stephenplusplus what are your thoughts?

stephenplusplus:
Sure, that would be fine. If google-auto-auth is invalidated, it should be chopped!

PR #78, opened 2018-04-04
chore: es6ify service.ts

### LINKED ISSUE #86: Use es6 classes and exports in service.ts

JustinBeckwith:
Resolves #86


src/service.ts
```diff
@@ -55,175 +60,171 @@ interface ServiceOptions {
...
- *
- * @param {object} config - Configuration object.
- * @param {string} config.baseUrl - The base URL to make API requests to.
- * @param {string[]} config.scopes - The scopes required for the request.
- * @param {object=} options - [Configuration object](#/docs).
- */
-function Service(config: ServiceConfig, options?: ServiceOptions) {
-  options = options || {};
-
-  util.privatize(this, 'baseUrl', config.baseUrl);
```
JustinBeckwith: I'm not 100% what privatize does.  I tried making most of these fields private, but they ended up being used by the unit tests so I had to make them public.  My kingdom for `internal`. @ofrobots @stephenplusplus any ideas here?

stephenplusplus: `privatize` was a new addition to resolve #13. In JS-land, it would just "hide" properties from `console.log`. I'm not sure what the modern equivalent is.

callmehiphop: Pretty sure that native class getters are not enumerable as well. Although I'm not sure if that's actually useful in this scenario.

src/service.ts
```diff
@@ -55,175 +62,177 @@ interface ServiceOptions {
...
+   */
+  getProjectId(): Promise<string>;
+  getProjectId(callback: (err: Error|null, projectId?: string) => void): void;
+  getProjectId(callback?: (err: Error|null, projectId?: string) => void): Promise<string>|void {
+    if (!callback) {
+      return this.getProjectIdAsync();
     }
+    this.getProjectIdAsync()
+        .then(p => callback(null, p))
+        .catch(e => callback(e));
```
kjin: If `callback` throws in `then` it will be called a second time in the `catch` handler. We should use `.then(p => callback(null, p), e => callback(e))`

src/service.ts
```diff
@@ -55,175 +62,177 @@ interface ServiceOptions {
...
+  }
 
-    callback(null, self.projectId);
-  });
-};
+  protected async getProjectIdAsync(): Promise<string> {
+    const projectId = await pify(this.authClient.getProjectId)();
```
kjin: Maybe we should cache `pify(this.authClient.getProjectId)`?

src/service.ts
```diff
@@ -55,175 +62,177 @@ interface ServiceOptions {
...
+  }
 
-    if (self.projectId === PROJECT_ID_TOKEN && projectId) {
-      self.projectId = projectId;
+  /**
+   * Get and update the Service's project ID.
+   *
+   * @param {function} callback - The callback function.
+   */
+  getProjectId(): Promise<string>;
```
kjin: Is this adding Promise functionality to `getProjectId`? If so maybe this change should be a `feat`?

src/service.ts
```diff
@@ -55,175 +62,177 @@ interface ServiceOptions {
...
+
+  private baseUrl: string;
+  private globalInterceptors;
+  private interceptors: ({ request(opts: r.Options): r.Options})[];
+  private packageJson: PackageJson;
+  private projectId: string;
+  private projectIdRequired: boolean;
+  private Promise: Promise<{}>;
+  // TODO: make this private
+  makeAuthenticatedRequest;
```
kjin: Can you add types to these objects? Looking at this file itself I'm not sure if it can potentially return a `Promise` or not (I sometimes see `return this.makeAuthenticatedRequest(opts, callback)` which seems strange to me)

src/service.ts
```diff
@@ -55,175 +62,177 @@ interface ServiceOptions {
...
+  /**
+   * Make an authenticated API request.
+   *
+   * @private
+   *
+   * @param {object} reqOpts - Request options that are passed to `request`.
+   * @param {string} reqOpts.uri - A URI relative to the baseUrl.
+   * @param {function} callback - The callback function passed to `request`.
+   */
+  private request(reqOpts: ExtendedRequestOptions, callback: (err: Error|null) => void) {
```
kjin: Should this be `protected`? I thought `Service` was meant to be subclassed.

src/service.ts
```diff
@@ -55,175 +62,177 @@ interface ServiceOptions {
...
+  /**
+   * Make an authenticated API request.
+   *
+   * @private
+   *
+   * @param {object} reqOpts - Request options that are passed to `request`.
+   * @param {string} reqOpts.uri - A URI relative to the baseUrl.
+   * @param {function} callback - The callback function passed to `request`.
+   */
+  request_(reqOpts: r.Options & ExtendedRequestOptions, callback?: (err: Error|null) => void) {
```
kjin: Mark this `private`?

src/service.ts
```diff
@@ -55,175 +62,177 @@ interface ServiceOptions {
...
-};
+  /**
+   * Make an authenticated API request.
+   *
+   * @private
+   *
+   * @param {object} reqOpts - Request options that are passed to `request`.
+   * @param {string} reqOpts.uri - A URI relative to the baseUrl.
+   */
+  private requestStream(reqOpts: ExtendedRequestOptions) {
```
kjin: Same as `request` comment

test/index.ts
```diff
@@ -15,37 +15,16 @@
...
-    assert.deepEqual(common, {
-      logger: fakeLogger,
-      Operation: fakeOperation,
-      paginator: fakePaginator,
-      Service: fakeService,
-      ServiceObject: fakeServiceObject,
-      util: fakeUtil,
-    });
+    assert(common.logger);
+    assert(common.logger);
```
kjin: Duplicate line

test/service.ts
```diff
@@ -17,9 +17,12 @@
 import * as assert from 'assert';
 import * as extend from 'extend';
 import * as proxyquire from 'proxyquire';
+import {Service, ServiceConfig} from '../src/service';
```
kjin: Maybe make this `import {Service as ServiceClass}` and then later `const Service: typeof ServiceClass = proxyquire('...')`

src/service.ts
JustinBeckwith: Woah, TIL.  If I use the rejected handler in here, should I also have a `.catch`?

src/service.ts
JustinBeckwith: This is going to be replaced with google-auth-library in #53, and it does async naturally.  I'd like to keep this simple before we do that change (hopefully next weekish)

src/service.ts
kjin: I don't think there needs to be a `.catch` since it's user code that's would be throwing.

src/service.ts
```diff
@@ -55,175 +62,178 @@ interface ServiceOptions {
...
+   */
+  getProjectId(): Promise<string>;
+  getProjectId(callback: (err: Error|null, projectId?: string) => void): void;
+  getProjectId(callback?: (err: Error|null, projectId?: string) => void): Promise<string>|void {
+    if (!callback) {
+      return this.getProjectIdAsync();
     }
+    this.getProjectIdAsync()
+        .then(p => callback(null, p), e => callback(e))
+        .catch(e => callback(e));
```
jinwoo: Sorry for the late comment, but my static analysis tool emitted an error when I ran it against this repository.
```
Rejection must be handled from .then().: "this.getProjectIdAsync()
        .then(p => callback(null, p), e => callback(e))
        .catch(e => callback(e))": /Users/jinwoo/src/nodejs-common/src/service.ts (141, 5)
```

With the current code, `callback` can be called twice if it ever throws some error.

This is what @ofrobots came up with as a bad code pattern and I added it to my analysis tool.

PR #92, opened 2018-04-08
test: limit codecov reporting

JustinBeckwith:
This limits the scope of codecov reporting to the src directory.


JustinBeckwith:
Ahhh, yeah we should just edit the .nycrc (sorry I missed that).  Will wait to hear back until I figure out what to do with this PR :)

PR #93, opened 2018-04-09
chore: remove through2

JustinBeckwith:
I ain't no expert in nodejs streams. But at a glance, it didn't look like we really needed `through2` in the mix here.  What do y'all think?  Does this PR pretty much do what through2 was doing for us?


stephenplusplus:
This should be good 👍

PR #95, opened 2018-04-10
chore: convert operation.ts to es6

### LINKED ISSUE #81: Use es6 classes and exports in operation.ts
Lets upgrade operation.ts to use `export class` syntax instead of `export = `.

Blocked by #89

### LINKED ISSUE #84: Remove `modelo` npm module
This module adds multiple inheritance, and probably needs to get removed. 

Blocked by #89

JustinBeckwith:
Resolves #81 
Resolves #84

### DIFF
--- a/src/service-object.ts
+++ b/src/service-object.ts
@@ -22,6 +22,7 @@ import * as arrify from 'arrify';
 import * as extend from 'extend';
 import * as is from 'is';
 import * as r from 'request';
+import { EventEmitter } from 'events';
 
 /**
  * @type {module:common/util}
@@ -49,12 +50,12 @@ export interface ServiceObjectConfig {
   /**
    * The base URL to make API requests to.
    */
-  baseUrl: string;
+  baseUrl?: string;
 
   /**
    * The method which creates this object.
    */
-  createMethod: Function;
+  createMethod?: Function;
 
   /**
    * The identifier of the object. For example, the name of a Storage bucket or Pub/Sub topic.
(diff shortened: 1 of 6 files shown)


src/service-object.ts
```diff
@@ -49,12 +50,12 @@ export interface ServiceObjectConfig {
...
    * The base URL to make API requests to.
    */
-  baseUrl: string;
+  baseUrl?: string;
 
   /**
    * The method which creates this object.
    */
-  createMethod: Function;
+  createMethod?: Function;
```
kjin: Is `createMethod` really optional? What if `create` is called if `createMethod` isn't defined?

stephenplusplus: The extending class, e.g. `Bucket` from the Storage API, defines what methods it needs in the config: https://github.com/googleapis/nodejs-storage/blob/1837107d425cd2da6051598a6779418baa6a7a6f/src/bucket.js#L65

If it doesn't provide `create`, then it's not on the object: https://github.com/JustinBeckwith/nodejs-common/blob/425bbfd1ae76dd54e91a80bff362a02deab41b68/src/service-object.ts#L168.

JustinBeckwith: In my case, I just noticed (well, got a compiler error) because the tests didn't pass one in.  That made me assume it's actually optional :)

kjin: Ah, that seems pretty tricky to pin down with types.

ofrobots:
Why is @wafflebot adding the cla label?

JustinBeckwith:
This is a good question for @lukesneeringer 🤣

PR #98, opened 2018-04-11
feat: convert util.ts to es6

### LINKED ISSUE #85: Use es6 classes and exports for util.ts

JustinBeckwith:
Resolves #85

### DIFF
--- a/src/util.ts
+++ b/src/util.ts
@@ -37,7 +37,16 @@ const request = r.defaults({
   },
 });
 
-export interface PromisifyAllOptions {
+export interface MakeAuthenticatedRequest {
+  (reqOpts: DecorateRequestOptions, optionsOrCallback?: MakeAuthenticatedRequestOptions|OnAuthenticatedCallback): void|Abortable|duplexify.Duplexify;
+  getCredentials: Function;
+  authClient: any;
+}
+
+export type Abortable = { abort(): void };
+export type AbortableDuplex = duplexify.Duplexify & Abortable;
+
+export interface PromisifyAllOptions extends PromisifyOptions {
   /**
    * Array of methods to ignore when promisifying.
    */
@@ -122,11 +131,15 @@ export interface MakeAuthenticatedRequestFactoryConfig {
 
   projectId?: string;
 
-  stream?: Stream;
+  stream?: duplexify.Duplexify;
 }
 
 export interface MakeAuthenticatedRequestOptions {
+  onAuthenticated: OnAuthenticatedCallback;
+}
 
+export interface OnAuthenticatedCallback {
+  (err: Error|null, reqOpts?: DecorateRequestOptions): void;
 }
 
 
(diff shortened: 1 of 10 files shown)


src/util.ts
```diff
@@ -127,854 +236,704 @@ class PartialFailureError extends Error {
...
+   * Maximum number of automatic retries attempted before returning the error. (default: 3)
+   */
+  maxRetries?: number;
 
-  return parsedHttpRespBody;
+  stream?: duplexify.Duplexify;
 }
 
-util.parseHttpRespBody = parseHttpRespBody;
+export class Util {
```
kjin: Initial thought -- I don't know that making `Util` an instantiable object makes sense, since util files are generally just collections of helper functions. I think we should be exporting each function individually instead, so other files only import the functions they need.

JustinBeckwith: That's a great question!  This was my initial approach.  When I did that, it broke pretty much all of the tests.  The tests are mocking `util` in a way that depends on the same function reference being stubbed as one invoked by the function under test.  Essentially - unless there's an object instance that can have it's methods stubbed reliably, none of the tests will come close to working.

It's possible I'm missing something, but this felt like a reasonable first step in the transition.

kjin: You can use `shimmer` to wrap/mock functions directly exported by a module. This will work:
```ts
import * as shimmer from 'shimmer';
import { privatize } from '../src/util';
import * as util from '../src/util';

shimmer.wrap(util, 'privatize', () => {
  return () => 'foo';
});

console.log(privatize({}, '', '')); // prints 'foo'
```
However, it won't work when mocking a function exported by `util` to test a different function exported by `util` (in other words, if `util` exports a function that calls another function in `util`).

Does that work for the cases you described?

JustinBeckwith: > However, it won't work when mocking a function exported by util to test a different function exported by util (in other words, if util exports a function that calls another function in util).

This is *exactly* what I was running into, and why I had to use the instance approach :(

src/util.ts
```diff
@@ -37,43 +37,153 @@ const request = r.defaults({
   },
 });
 
-const util = module.exports;
+export interface PromisifyAllOptions {
```
kjin: Maybe `PromisifyAllOptions` can extend `PromsifyOptions`? That way the `as any` in `promisifyAll` is unneeded.

JustinBeckwith:
@kjin no more `as any` 🎆

src/util.ts
```diff
@@ -745,9 +776,9 @@ export class Util {
    * @param {object} localConfig - Service-level configurations.
    * @return {object} config - Merged and validated configuration.
    */
-  normalizeArguments(globalContext: GlobalContext, localConfig: {}) {
-    const globalConfig = globalContext && globalContext.config_;
-    return util.extendGlobalConfig(globalConfig as any, localConfig as any);
+  normalizeArguments(globalContext: GlobalContext, localConfig: GlobalConfig) {
+    const globalConfig = globalContext && globalContext.config_ as GlobalConfig;
```
kjin: Nit: Can `GlobalContext.config_` be of type `GlobalConfig`?

PR #106, opened 2018-04-17
chore: round 2 of gts fixing

JustinBeckwith:


JustinBeckwith:
@stephenplusplus @googleapis/node-team someone mind taking a gander?

src/operation.ts
```diff
@@ -76,8 +78,8 @@ export class Operation extends ServiceObject {
    * @return {promise}
    */
   promise() {
-    return new this.Promise(
-        (resolve: Function, reject: (err: Error) => void) => {
+    return new this.Promise!
```
ofrobots: Not a comment for this PR, but in the next semver major, can we drop support from custom `Promise`s? Note that async/await _always_ use native promises, so the use-case for custom promises is growing thin, and might even be footgun as things behave differently from what people might expect.

Is there a strong reason to continue supporting this? People can always do `global.Promise = that.thing`;

PR #112, opened 2018-04-23
chore: make internal methods async

JustinBeckwith:


src/service-object.ts
```diff
@@ -329,16 +336,14 @@ class ServiceObject extends EventEmitter {
...
-      if (err) {
-        callback(err, null, resp);
-        return;
-      }
-
-      self.metadata = resp as Metadata;
-
-      callback(null, self.metadata, resp);
-    });
+    this.request(reqOpts).then(
```
ofrobots: Likewise, here and elsewhere, can we convert the containing method to async, and use `await`?

JustinBeckwith: In this case we can't.  `getMetadata` is a public function, and I don't want to make this change just yet.

src/service-object.ts
```diff
@@ -409,11 +415,11 @@ class ServiceObject extends EventEmitter {
 
     reqOpts.interceptors_ = childInterceptors.concat(localInterceptors);
 
-    if (!callback) {
+    if (returnStream) {
```
stephenplusplus: Just a variable naming preference you are free to disregard. `shouldReturnStream` indicates to me it's a boolean, where as `returnStream` makes me think it's a Stream that we're returning.

stephenplusplus: And one more thought-- to be more idiomatic, can it either be part of `reqOpts` or another object for configuration? Seeing `this.request_(reqOpts, true)` is a bit too Java-- even though smart IDEs will get it, it can be troubling for GH-browsing.

JustinBeckwith: > is a bit too Java

Hey!  I'm a C# developer thank you.

PR #113, opened 2018-04-25
chore: unroll nested functions

JustinBeckwith:
So -  I don't feel great about this change.  The `makeAuthenticatedRequestFactor` stuff in `src/util.ts` is the primary touch point of `google-auto-auth`, and is next up on the replace list.  This function has 3 nested functions, which was making the code *really* hard for me to understand.  I took a shot at trying to flatten it... but I'm not sure I made this code any more readable.  

I guess I'm looking to use this PR as a way to start a conversation.  

1. How do we feel about nested functions?  I am personally 👎 on them, mostly for readability purposes. 

2. In this specific situation, how do we make this beast a little less unwieldy? 

3. In the general sense, does it make sense to keep the `makeAuthenticatedRequestFactory` method around? Using `GoogleAuth.request()`, `google-auth-library` just kinda...  does it for you.  I don't know which other scenarios this method is helping to support. 

Thoughts appreciated :)

### DIFF
--- a/src/util.ts
+++ b/src/util.ts
@@ -20,6 +20,7 @@
 
 import * as duplexify from 'duplexify';
 import * as ent from 'ent';
+import {EventEmitter} from 'events';
 import * as extend from 'extend';
 import * as is from 'is';
 import * as r from 'request';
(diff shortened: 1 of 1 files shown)


callmehiphop:
Definitely into the idea of using the auth clients request functionality, I think we'll still need a small wrapper for retrying requests (I couldn't find any docs in axios claiming it supported that kind of functionality). But I imagine as a result it would require a lot less code on our end and we could finally drop `request` as a dependency.

stephenplusplus:
Good call on the retry part. I updated my list to mention retry-request isn't required, as long as it's replaced by something that functions the same.

PR #114, opened 2018-04-27
feat: google-auto-auth ==> google-auth-library

### LINKED ISSUE #53: Switch to google-auth-library

JustinBeckwith:
Resolves #53.

@stephenplusplus @alexander-fenster the system-tests folder in this library is noticeably absent :) I'm kinda scared to submit this without doing some sanity testing.  Is there a specific veneer I should test this with?  Something that really exercises all the things?


alexander-fenster:
There are no system tests indeed. I would suggest that you can check spanner, bigquery, and pubsub, and if their system tests work with your changes, everything else will likely work. If you make them depend on `"@google-cloud/common": "JustinBeckwith/nodejs-common#promisy"` it will take it right from here.

Also, one unit test fails :)

alexander-fenster:
An easy way to try system tests for spanner, pubsub, etc. is making a branch in their repos (not forking a repo), editing `package.json` in that branch, and also removing master branch filter in `.circleci/config.yml`. That way it will run system and samples tests in your branch and you'll save time on setting up keys and doing all that stuff locally.

src/util.ts
```diff
@@ -30,7 +31,7 @@ import * as streamEvents from 'stream-events';
 import {Service} from '.';
 import {Interceptor} from './service-object';
 
-const googleAuth = require('google-auto-auth');
+
```
alexander-fenster: nit: remove empty line

package.json
```diff
@@ -43,6 +43,7 @@
     "samples-test": "cd samples/ && npm link ../ && npm test && cd ../",
     "test-no-cover": "repo-tools test run --cmd mocha -- build/test/*.js --no-timeouts",
     "test": "repo-tools test run --cmd npm -- run cover",
+    "test-docker": "docker build . -t test && docker run -it test",
```
alexander-fenster: hmm, looks like this should be in a separate pr, shouldn't it?

package.json
```diff
@@ -53,10 +54,11 @@
   },
   "dependencies": {
     "arrify": "^1.0.1",
+    "axios": "^0.18.0",
```
alexander-fenster: should it be a devDependency?

JustinBeckwith: It's already getting pulled in by `google-auth-library`, and it's going to be more directly used in #115.  I'm ok with it being in the regular deps :)

package.json
JustinBeckwith: Fair point :)

PR #125, opened 2018-05-01
fix: back-compat issues with cloud-profiler

### LINKED ISSUE #123: Problems migrating to v0.18.X
Right now, we're using ServiceObject.request() with a callback so we can get the error, body, and response body even when there's an error. We need to do this because the profiler API will return an error code, with a backoff specified in the body.

What can we do to migrate?

JustinBeckwith:
Resolves #123

### DIFF
--- a/src/service-object.ts
+++ b/src/service-object.ts
@@ -434,13 +434,12 @@ class ServiceObject extends EventEmitter {
   request(reqOpts: DecorateRequestOptions, callback: r.RequestCallback): void;
   request(reqOpts: DecorateRequestOptions, callback?: r.RequestCallback):
       void|Promise<r.Response> {
-    if (callback) {
-      this.request_(reqOpts).then(
-          r => callback(null, r as r.Response, r.body),
-          e => callback(e, null!, null));
-      return;
+    if (!callback) {
+      return this.request_(reqOpts) as Promise<r.Response>;
     }
-    return this.request_(reqOpts) as Promise<r.Response>;
+    this.request_(reqOpts).then(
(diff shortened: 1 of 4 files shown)


JustinBeckwith:
@nolanmar511 this should fix what you're running into :)

src/service-object.ts
```diff
@@ -430,7 +430,16 @@ class ServiceObject extends EventEmitter {
    * @param {string} reqOpts.uri - A URI relative to the baseUrl.
    * @param {function} callback - The callback function passed to `request`.
    */
-  request(reqOpts: DecorateRequestOptions): Promise<r.Response> {
+  request(reqOpts: DecorateRequestOptions): Promise<r.Response>;
+  request(reqOpts: DecorateRequestOptions, callback: r.RequestCallback): void;
+  request(reqOpts: DecorateRequestOptions, callback?: r.RequestCallback):
+      void|Promise<r.Response> {
+    if (callback) {
+      this.request_(reqOpts).then(
```
ofrobots: infinitesimally teeny nit. If you reverse the direction of this branch, you will need one less `return`.

src/service-object.ts
```diff
@@ -430,7 +430,16 @@ class ServiceObject extends EventEmitter {
...
+          r => callback(null, r as r.Response, r.body),
```
ofrobots: This line has a variable named `r` and a type named `r`. Can you remove the ambiguity even if the compiler / linter isn't complaining about it.

JustinBeckwith:
Added a test 🎉

PR #132, opened 2018-05-08
fix: repair makeAuthenticatedRequest callback order

### LINKED ISSUE #131: Service.request throws
Service.request throws an error:

```js
TypeError: Cannot read property 'body' of undefined
    at (...)/Documents/cloud-trace-nodejs/node_modules/@google-cloud/common/src/service.ts:227(...)
    at <anonymous>
    at process._tickCallback (internal/process/next_tick.js:188:7)
```

The line in question is:

https://github.com/googleapis/nodejs-common/blob/v0.18.4/src/service.ts#L227

My theory is that it's trying to read the `body` property from a field that was already a `body` property of another object. In other words, I believe this is a case where `(err, body, resp)` and `(err, resp, body)` has been mixed up. This is because tracing back to where the undefined object was born yields here:

https://github.com/googleapis/nodejs-common/blob/v0.18.4/src/util.ts#L323-L327

I'd love to help fix it but I'm not exactly sure what `makeAuthenticatedRequest` is doing. This section seems to pass a callback of `type OnAuthenticatedCallback = (err: Error|null, reqOpts?: DecorateRequestOptions) => void` when it really expects a `type RequestCallback = (err: Error|null, body: ResponseBody, res: r.Response) => void`, and the compiler seems to be OK with that (possible because `type ResponseBody = any`):

https://github.com/googleapis/nodejs-common/blob/v0.18.4/src/util.ts#L602-L603

I think ultimately that the compiler didn't catch this is the culprit, but I'm not sure what the correct thing should be.

#### Environment details

  - OS: macOS
  - Node.js version: 8
  - npm version: 5
  - @google-cloud/common version: 0.18.4

#### Steps to reproduce

  1. In the `cloud-trace-nodejs` repo: `npm install @google-cloud/common@0.18.4` followed by `npm run compile`, then run `$(npm bin)/mocha build/test/test-config-credentials.js`

JustinBeckwith:
I think? this resolves #131.  @kjin would you mind taking this branch for a spin?

### DIFF
--- a/src/service.ts
+++ b/src/service.ts
@@ -209,9 +209,18 @@ export class Service {
     });
 
     if (reqOpts.shouldReturnStream) {
-      return this.makeAuthenticatedRequest(reqOpts) as r.Request;
+      // tslint:disable-next-line:no-any
+      return this.makeAuthenticatedRequest(reqOpts) as any as r.Request;
     } else {
-      return pify(this.makeAuthenticatedRequest)(reqOpts);
+      return pify(this.makeAuthenticatedRequest, {multiArgs: true})(reqOpts)
+          .then(args => {
+            /**
+             * Note: this returns an array of results in the form of a
+             * BodyResponseCallback, which means: [body, response].  Return the
+             * response object in the promise result.
+             */
+            return args.length > 1 ? args[1] : null;
+          });
     }
   }
 
--- a/src/util.ts
+++ b/src/util.ts
@@ -266,7 +266,7 @@ export class PartialFailureError extends Error {
 }
 
 export interface BodyResponseCallback {
-  (err: Error|null, body: ResponseBody, res: r.Response): void;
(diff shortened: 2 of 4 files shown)


kjin:
Ok, that makes it a lot clearer, thanks.

In conjunction with this extra fix it would work:

https://github.com/kjin/nodejs-common/commit/6a5d88839ff1d482cb90d3322774917f089ecd87

JustinBeckwith:
*awesome*.  Thank you so much for helping hunt this down.  I added a few more type improvements, and integrated your fix.

JustinBeckwith:
Side note - this is one of the reasons pify is dangerous. You lose all typing, and it's easy to make mistakes like this one :(

JustinBeckwith:
*gentle ping* @stephenplusplus @kjin I'd like to get this merged so we can cut a release today :)

kjin:
Thanks Justin!

PR #135, opened 2018-05-09
fix: swap resp and body in request callback

### LINKED ISSUE #134: 0.18.5 is breaking @google-cloud/profiler.
This is the section of code which is now failing:
```
this.request(
          options,
          (err: Error, body: object, response: http.ServerResponse) => {
            try {
              const prof = responseToProfileOrError(err, body, response);
              this.logger.debug(
                  `Successfully created profile ${prof.profileType}.`);
              resolve(prof);
            } catch (err) {
              reject(err);
            }
          });
    });
```
The `body` now has the format:
```
{
    statusCode: 200,
    body:  {
         "name":"projects/projectId/XXXXX",
         "profileType":"WALL"
         .....
    }
   .........
```

We expect to see just what's in the `body` portion of that.

kjin:
Fixes #134

- [x] Tests and linter pass

### DIFF
--- a/src/service-object.ts
+++ b/src/service-object.ts
@@ -26,7 +26,7 @@ import * as r from 'request';
 import {Duplex} from 'stream';
 
 import {Service} from '.';
-import {ApiError, DecorateRequestOptions, util} from './util';
+import {ApiError, BodyResponseCallback, DecorateRequestOptions, util} from './util';
 
 export interface Interceptor {
   // tslint:disable-next-line:no-any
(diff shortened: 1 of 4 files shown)


JustinBeckwith:
why

kjin:
The callback passed to `Service#request` should be of type `BodyResponseCallback` but was `request.Response` type instead. 0.18.5 changed the logic to fit the type of the latter, when the type should've been changed to the former (but no matter what the type was, the fix made in #132 needed to be done anyway)

JustinBeckwith:
@kjin assuming you tried this out on something? The lack of system-tests in this repo is kinda scary.

kjin:
@JustinBeckwith I verified that it fixes #134 (when https://github.com/GoogleCloudPlatform/cloud-profiler-nodejs/pull/193 is applied). I'll also try it with nodejs-storage.

JustinBeckwith:
👌

PR #142, opened 2018-05-18
feat: remove `levels` from LoggerConfig and don't expose index property on Logger

kjin:
__edit:__ _In addition to the below, this PR removes `levels` from the documented API of `LoggerConfig`. There are no behavioral changes here, and impact should overall be minimal because `LoggerConfig` (along with `Logger`) was introduced in 0.18._

This change makes it so that object of type `Logger` by default expose log levels corresponding to the default options. This is the primary use case for objects of type `Logger`.

For `Logger` objects with custom log levels:

```ts
const l = new Logger() as CustomLevelsLogger;
l.baby('foo');
```

This makes mocking `Logger` objects easier, and also gives us code completion for default log levels.

### DIFF
--- a/src/logger-compat.ts
+++ b/src/logger-compat.ts
@@ -20,7 +20,7 @@
 
 import * as is from 'is';
 
-import {kFormat, Logger, LoggerConfig} from './logger';
+import {CustomLevelsLogger, kFormat, Logger, LoggerConfig} from './logger';
 
 // tslint:disable-next-line:no-any
 function isString(obj: any): obj is string {
@@ -48,7 +48,7 @@ function createLogger(optionsOrLevel?: Partial<LoggerConfig>|string) {
   });
   return Object.assign(
       // tslint:disable-next-line:no-any
-      result as Logger & {format: (...args: any[]) => string},
+      result as CustomLevelsLogger & {format: (...args: any[]) => string},
       {levels: options.levels, level: options.level});
 }
 
--- a/src/logger-compat.ts
+++ b/src/logger-compat.ts
@@ -20,23 +20,43 @@
 
 import * as is from 'is';
 
-import {CustomLevelsLogger, kFormat, Logger, LoggerConfig} from './logger';
+import {kFormat, LEVELS, Logger, LoggerConfig} from './logger';
+
+export interface CustomLevelsLoggerConfig extends LoggerConfig {
+  /**
+   * The list of levels to use.
+   */
+  levels: string[];
+}
+
+// tslint:disable:no-any
+export type CustomLevelsLoggerResult = {
+  [logLevel: string]: (...args: any[]) => CustomLevelsLoggerResult;
+}&{
+  format: (...args: any[]) => string;
+  levels: string[];
+  level: string|false;
+};
+// tslint:enable:no-any
 
 // tslint:disable-next-line:no-any
 function isString(obj: any): obj is string {
   return is.string(obj);
 }
 
-function createLogger(optionsOrLevel?: Partial<LoggerConfig>|string) {
+function createLogger(optionsOrLevel?: Partial<CustomLevelsLoggerConfig>|
+                      string): CustomLevelsLoggerResult {
   // Canonicalize input.
   if (isString(optionsOrLevel)) {
     optionsOrLevel = {
       level: optionsOrLevel,
     };
   }
-  const options: LoggerConfig =
-      Object.assign({}, Logger.DEFAULT_OPTIONS, optionsOrLevel);
-  const result = new Logger(options);
+  const options: CustomLevelsLoggerConfig =
+      Object.assign({levels: LEVELS}, Logger.DEFAULT_OPTIONS, optionsOrLevel);
+  // ts: We construct other fields on result after its declaration.
+  // tslint:disable-next-line:no-any
+  const result: CustomLevelsLoggerResult = new Logger(options) as any;
   Object.defineProperty(result, 'format', {
     get() {
       return result[kFormat];
@@ -46,13 +66,8 @@ function createLogger(optionsOrLevel?: Partial<LoggerConfig>|string) {
       result[kFormat] = value.bind(result);
     }
   });
-  return Object.assign(
-      // tslint:disable-next-line:no-any
-      result as CustomLevelsLogger & {format: (...args: any[]) => string},
-      {levels: options.levels, level: options.level});
+  return Object.assign(result, {levels: options.levels, level: options.level});
 }
-
-const LEVELS = Logger.DEFAULT_OPTIONS.levels;
 
 /**
  * Create a logger to print output to the console.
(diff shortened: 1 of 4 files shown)


src/logger.ts
```diff
@@ -53,16 +56,16 @@ export class Logger {
   static DEFAULT_OPTIONS:
       Readonly<LoggerConfig> = {level: 'error', levels: LEVELS, tag: ''};
 
-  // TODO: Mark this private when TypeScript 2.9 comes out.
-  // See https://github.com/Microsoft/TypeScript/issues/20080 for more
-  // information.
-  [kTag]: string;
+  private[kTag]: string;
```
jinwoo: Nit. A space between `private` and `[kTag]`.

src/logger.ts
```diff
@@ -53,16 +56,16 @@ export class Logger {
...
 
-  /**
-   * Emits a log at this log level.
-   */
-  // tslint:disable-next-line:no-any
-  [logLevel: string]: (...args: any[]) => this;
+  // tslint:disable:no-any
+  silent!: (...args: any[]) => this;
```
jinwoo: Why do you need the non-null annotations in these fields?

src/logger.ts
```diff
@@ -88,22 +91,35 @@ export class Logger {
...
+        // tslint:disable-next-line:no-any
+        (this as any)[level] = () => this;
       }
     }
   }
 
-  // TODO: Mark this as protected when TypeScript 2.9 comes out.
   // tslint:disable-next-line:no-any
-  [kFormat](level: string, ...args: any[]): string {
+  private[kFormat](level: string, ...args: any[]): string {
```
jinwoo: Nit. A space between `private` and `[kFormat]`. Or is this what clang-format does?

src/logger.ts
```diff
@@ -88,22 +91,35 @@ export class Logger {
     for (let i = 0; i < options.levels.length; i++) {
       const level = options.levels[i];
       if (i <= levelIndex) {
-        this[level] = (...args) => {
+        // ts: this doesn't have an index signature, but we want to set
+        // properties anyway.
+        // tslint:disable-next-line:no-any
+        (this as any)[level] = (...args: any[]) => {
```
jinwoo: Are `options.levels` limited to info, warn, error, debug, silent, and silly? Or is any string allowed? If the latter, I don't see the point of separating out `CustomLevelsLogger`.

src/logger.ts
kjin: This is what clang-format does.

src/logger.ts
kjin: The TypeScript compiler doesn't know that these will be set for default options. I'll add a note in the comments.

src/logger.ts
kjin: Separating out `CustomLevelsLogger` makes it impossible to mock or extend the logger in a useful way.

Using custom log levels is a use case that is not useful for any Google-authored module anyway, because they always use the defaults.

src/logger.ts
```diff
@@ -53,16 +56,19 @@ export class Logger {
...
-  /**
-   * Emits a log at this log level.
-   */
-  // tslint:disable-next-line:no-any
-  [logLevel: string]: (...args: any[]) => this;
+  // ts: The compiler can't statically detect that these will be definitely
+  // assigned. They will be if default log levels are used, so we use non-null
+  // annotations here.
+  // tslint:disable:no-any
+  silent!: (...args: any[]) => this;
```
jinwoo: I'm still not convinced that we should use these non-null annotations. If an instance of this class is created with custom levels, some (or any) of these won't be set, right? How do we prevent coding errors in that case?

kjin: The alternative, if we want to externalize log functions, is to also have a second `DefaultLevelsLogger` in addition to `CustomLevelsLogger`. This works but also means that any time a logger is constructed, it needs to be casted to one of the two (otherwise it has no public API). I've been trying to finagle with TypeScript to discover other alternatives (like constructor overloading) but there aren't any AFAIK.

Personally I think we should get rid of custom levels, as they are virtually unused in practice. For client libraries we don't use them -- and I don't see a reason for third-party modules to have a direct dependency on this module for its logger. Because the role of the `nodejs-common` module as a collection of helpers for GCP client libraries, I think that if someone is actually using `Logger` for other reasons, then they are on their own.

jinwoo: What about always adding default log levels regardless of whether they are included in the custom options?

kjin: That seems like a semver-major change. The library is experimental, so it's probably reasonable to do so (ping @JustinBeckwith) -- but I think if we were to make a semver-major change, we might as well remove custom log levels entirely instead.

kjin: I think though that using the default log levels should still be considered a coding error, regardless of whether they're added by default.

jinwoo: I thought the default log levels were supposed to be always there even though you were using custom log levels. I think I thought that way because the non-null operators were put to those default log functions.

In that case, I think we should have a new base class (say `BaseLogger`), from which both the `Logger` class and the `CustomLevelsLogger` class should inherit. `Logger` will have these logging functions by default while `BaseLogger` and `CustomLevelsLogger` won't. WDYT?

BTW if this PR blocks your work, I can LGTM as is.

kjin: That makes sense. I think I prototyped something similar when I was working on this PR -- I'll try it out again. (Though, to keep things stable, `Logger` will be the one that will stay the same, not `BaseLogger`.)

src/logger-compat.ts
```diff
@@ -20,23 +20,43 @@
...
     };
   }
-  const options: LoggerConfig =
-      Object.assign({}, Logger.DEFAULT_OPTIONS, optionsOrLevel);
-  const result = new Logger(options);
+  const options: CustomLevelsLoggerConfig =
+      Object.assign({levels: LEVELS}, Logger.DEFAULT_OPTIONS, optionsOrLevel);
+  // ts: We construct other fields on result after its declaration.
+  // tslint:disable-next-line:no-any
+  const result: CustomLevelsLoggerResult = new Logger(options) as any;
```
jinwoo: I think you can use `{}` instead of `any` here.

src/logger.ts
```diff
@@ -50,19 +48,20 @@ export class Logger {
...
 
-  /**
-   * Emits a log at this log level.
-   */
-  // tslint:disable-next-line:no-any
-  [logLevel: string]: (...args: any[]) => this;
+  // ts: The compiler can't statically detect that these will be definitely
+  // assigned, so we use non-null annotations here.
+  // tslint:disable:no-any
+  silent!: (...args: any[]) => this;
```
jinwoo: Can these be now defined as regular methods, by introducing a private method?

```ts
private logHelper(level: string, ...args: any[]) {
  args.unshift(level);
  console.log(this[kFormat].apply(this, args));
  return this;
}

silent(...args: any[]) {
  return this.logHelper('silent', ...args);
}
...
```

src/logger-compat.ts
kjin: I tried it and it doesn't work, probably because `{}` cannot be casted to `CustomLevelsLoggerResult`.

Issue #154, opened 2018-05-31
0.19.0 is empty 😭


JustinBeckwith:
It looks like the `npm prepare` script failed to actually run:
https://circleci.com/gh/googleapis/nodejs-common/2847

PR #155, opened 2018-05-31
chore: run CI as non-root user

alexander-fenster:
`npm publish` did not work when run as root. Let's run as user `node` as we do in all other projects.

- [x] Tests and linter pass

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -52,6 +52,7 @@ jobs:
   node4:
     docker:
       - image: 'node:4'
+        user: node
     steps: &unit_tests_steps
       - checkout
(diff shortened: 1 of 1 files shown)


alexander-fenster:
Fixes #154.

JustinBeckwith:
It looks like this was introduced as part of https://github.com/googleapis/nodejs-common/pull/147, which looks to be an automated change.  Do we need to fix this in the template and then make the change in all the CircleCI configs?

alexander-fenster:
I was under impression it's already in all CI configs - I changed all (well, turns out like not all but most) to run under user node long time ago. Well, I screwed up.

PR #157, opened 2018-05-31
feat: expose Logger.LEVELS

kjin:
This change just puts `LEVELS` as a static field on the `Logger` object.

### DIFF
--- a/src/logger.ts
+++ b/src/logger.ts
@@ -33,11 +33,6 @@ export interface LoggerConfig {
   tag: string;
 }
 
-/**
- * The default list of log levels.
- */
-export const LEVELS = ['silent', 'error', 'warn', 'info', 'debug', 'silly'];
-
 export const kFormat = Symbol('Logger formatter');
 export const kTag = Symbol('Logger tag format');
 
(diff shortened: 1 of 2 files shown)


src/logger.ts
```diff
@@ -49,6 +44,10 @@ export class Logger {
    * Default logger options.
    */
   static DEFAULT_OPTIONS: Readonly<LoggerConfig> = {level: 'error', tag: ''};
+  /**
+   * The default list of log levels.
+   */
+  static LEVELS = ['silent', 'error', 'warn', 'info', 'debug', 'silly'];
```
jinwoo: Why is this better than the module-level definition? I prefer the latter personally.

And if you want to do this, can you put `readonly` here? Also `readonly` for `DEFAULT_OPTIONS` above (possibly `private` too there?).

kjin: If we expose them at the module level we need to rename it, because otherwise it would place `LEVELS` at the top-level with `Logger`, `Service`, etc. Regardless of whether we rename it, I still think it makes sense to nest it under `Logger` for the same reason -- it shouldn't be a top-level export.

jinwoo: I'm not sure what you mean by "we need to rename it". This is fine. I don't feel too strongly about module-level const vs class static anyway.

jinwoo: You could rename it `LOG_LEVELS` when re-exporting it from `index.ts` :)

PR #162, opened 2018-06-07
chore: upgrade to latest of all dependencies

JustinBeckwith:


test/util.ts
```diff
@@ -1761,14 +1755,15 @@ describe('common/util', () => {
     // in node v4 which is why the eval call is wrapped in a try catch block.
     try {
       eval(`
+        const assert2 = require('assert');
```
ofrobots: Why require assert again here?

src/service.ts
```diff
@@ -18,13 +18,13 @@
...
-import * as arrify from 'arrify';
+import arrify from 'arrify';
 import {Duplexify} from 'duplexify';
-import * as extend from 'extend';
+import extend from 'extend';
 import {GoogleAuth} from 'google-auth-library';
-import * as is from 'is';
-import * as pify from 'pify';
-import * as r from 'request';
+import is from 'is';
```
ofrobots: not a comment for this PR, but if possible, we should switch to `@sindresorhus/is`. I find that `is` has non-intuitive behavior, e.g. `is.object(new Error())` returns false.

JustinBeckwith: :nods: I opened #163 to track

test/util.ts
JustinBeckwith: There's some weirdness with the combination of synthetic imports and the eval.  I get a `ReferenceError: assert is not defined` if I don't re-import it using `require`.  Not entirely sure what's happening.

JustinBeckwith:
*gentle ping* @alexander-fenster @stephenplusplus @ofrobots

alexander-fenster (approved):

PR #167, opened 2018-06-16
feat: export auth and GoogleAuth

JustinBeckwith:
Trying to reduce the # of diamond dependencies in our own modules.


ofrobots (changes requested):
I would prefer an auth *instance* was exported on `Service` that consumers would use rather than instantiating their own auth with possibly different config. For example, `Service` already exposes `getProjectId`.

What are the use-case for consumer libraries to create their own auth instances? I am going to block this until I better under the use-cases.

JustinBeckwith:
Great feedback.  To be entirely honest, I haven't dug that deep quite yet.  What I'm observing right now is that we import and use varying versions of `google-auth-library` across multiple libraries, and then pass objects back and forth.  I'm trying to create a choke point of sorts where *one* base library takes the dependency, and then it comes in with common.  Let's use [nodejs-spanner](https://github.com/googleapis/nodejs-spanner) as an example.

- [google-auto-auth](https://github.com/googleapis/nodejs-spanner/blob/master/package.json#L74) is brought in directly to the module.
- [nodejs-common](https://github.com/googleapis/nodejs-common/blob/master/package.json#L58) takes a dependency here.
- [google-gax](https://github.com/googleapis/gax-nodejs/blob/master/package.json#L15) brings it in as well

All of these modules are brought in as top level dependencies in `nodejs-spanner`, and auth objects and config are passed in and out of them.  

As a result - right now I need to update `google-gax`, `nodejs-common`, and `google-auth-library` all at the same time.  This feels hyper risky, and I was hoping to detangle things just a bit :)

Now onto your original question 🤣 Google auto auth is used in `nodejs-spanner` like this:
https://github.com/googleapis/nodejs-spanner/blob/master/src/index.js#L195

Today the top level `Spanner` class is responsible for creating the `GoogleAuth` object, and then setting that on the base `Service` class object.  I'm not really sure today how I could make this work without at least exporting the `GoogleAuth` class :(

stephenplusplus:
> google-auto-auth is brought in directly to the [nodejs-spanner] module.

We should be able to get rid of that. Every common.Service instance provides an [auth client instance](https://github.com/googleapis/nodejs-common/blob/0d1075b86939206d04053cf7108ed6ae372a6731/src/service.ts#L109), which we can use from Spanner.

> nodejs-common takes a dependency [on GoogleAuthLibrary] here.
> google-gax brings it in as well

I don't think flattening these will work, as they're kind of in competition with each other. An old plan was to transition entirely from CommonGrpc to GAX. But for now, both the common modules and GAX need to do their own auth, and neither depend on each other.

ofrobots:
@JustinBeckwith  The code you pointed to in `nodejs-spanner` is actually a good example making my point.

Ideally [this code](https://github.com/googleapis/nodejs-spanner/blob/a7e19825dffeaedcd4be03c1332a9749e8d56d19/src/index.js#L186-L195) should be lifted into the `Service` constructor instead of each `Service` subclass duplicating the work (and duplicating the dependency): 
```ts
  this.options = extend(
    {
      libName: 'gccl',
      libVersion: require('../package.json').version,
      scopes: scopes,
    },
    options
  );

  this.auth = googleAuth(this.options);
```

This code is generic. It has nothing to do do with `spanner` specifically.

I still don't see a valid reason for child libraries to directly need access to the auth client. Superclass facilities in either `common` or `gax` should handle auth construction for everyone. Expose auth directly (as this PR proposes) – while eliminating the dependency from child libraries – does not do enough to get rid of the code duplication and the technical debt.

stephenplusplus:
> I still don't see a valid reason for child libraries to directly need access to the auth client.

In Storage, we need to use the "sign blob" endpoints for signed policies and URLs.

In any gapic client, we need to get a project ID before we send a request to gapic.

ofrobots:
> In any gapic client, we need to get a project ID before we send a request to gapic.

This should be done via something like `await this.getProjectId();`. This can already be done in subclasses of `Service`.

> In Storage, we need to use the "sign blob" endpoints for signed policies and URLs.

It would be better to expose an auth *instance* rather than the ability to construct auth clients.

stephenplusplus:
> This should be done via something like await this.getProjectId();. This can already be done in subclasses of Service.

Yep, and that's fine. I'd say it's just as useful, if not better, to have the entirety of the auth instance, though.

Note that I was giving those examples as a response to your comment: "I still don't see a valid reason for child libraries to directly need access to the auth client." Judging from the comment I quote below, maybe you meant "auth module". And foreshadowing my response, I agree.

> It would be better to expose an auth instance rather than the ability to construct auth clients.

Agree (told you I would agree), and we do. Spanner was just not doing it right, it seems. I haven't gone in to fix it yet, so there's a chance it's required for some reason, but as far as I can tell, that was just a mistake.

ofrobots:
@stephenplusplus :+1:

PR #182, opened 2018-06-29
fix: ensure request returns response in error

JustinBeckwith:


src/service-object.ts
```diff
@@ -450,7 +450,8 @@ class ServiceObject extends EventEmitter {
     }
     this.request_(reqOpts).then(
         res => callback(null, res.body, res as r.Response),
-        err => callback(err, null, null!));
+        err => callback(
```
ofrobots: OMG LOL at the `null!`.

PR #183, opened 2018-06-29
test: test the typescript install

JustinBeckwith:
This ensures we don't leak synthetic imports to other TypeScript consumers.


tsconfig.json
```diff
@@ -12,6 +12,7 @@
     "test/**/*.ts"
   ],
   "exclude": [
-    "node_modules"
+    "node_modules",
+    "test/fixtures"
```
ofrobots: Another way doing this would have been to drop the `test/**/*.ts` from the `includes` above, but this is fine.

PR #189, opened 2018-07-09
feat: split the packages

JustinBeckwith:
Well this one was a hoot.  As I looked around, I started to notice a lot of our modules only use small pieces of what's available in nodejs-common.  The package is bringing `request` into everything else, when in many cases, it's just not needed.   To help make things easier to rationalize, this PR splits common into 4 different modules:
- paginator
- promisify
- projectify
- common

I used lerna to tie it all together.  Still have no idea how the whole circleci thing gonna work out, but we'll see where it goes :)

Thoughts?


stephenplusplus:
I like the idea of splitting into multiple modules 👍. My only question is, I'm not sure of the distinction. For example, why not split out the `Operation` class? Were the ones you chose just the most independent, "module-like" pieces?

JustinBeckwith:
So far I've only pulled out the ones that I've seen used independently.  Vision only uses promisify, pubsub only uses paginator, firestore only uses projectify, etc.  It's entirely possible we run into one where Operation is needed independently.  When we get there, I'm happy to factor it out too :)

ofrobots:
Yay for modularity!

For the record, token pasting and removal that necessitates `projectify` is a bug, and needs to be fixed – removing the need for the module. I do get it that at this point in time we are not there yet.

alexander-fenster:
Monorepo for the win! But I don't really understand how `lerna publish` is going to work in Circle :-(  (I'm sure it must be possible but I don't really see how exactly for now)

ofrobots:
I too would like some more thought to how to the release process would work. @kjin may have some thoughts about this OpenCensus for Node also started with learna. AFAIK, there is no answer yet on how releases are supposed to work.

JustinBeckwith:
For now, I think the process would be a human running `lerna publish`

JustinBeckwith:
@ofrobots @theacodes  I don't really know what to do here.  We all agree this stuff should be split up, but I'm not sure of a way to automate the releases.  Running `lerna publish` from a developer machine followed by a `git push` to push the new tags and such will work, but has the obvious disadvantages.

I see a few options here:
1. We drop the mono-repo, and create new repos for utility.  This isn't hard, it's just a pain. 
1. We use https://github.com/atlassian/lerna-semantic-release to do this work for us, but it requires adding the `GITHUB_TOKEN` to CircleCI env vars.  I don't think this is that bad, but I don't think it's policy atm.
1. We could add support for sub packages to `releasetool`. 
1. We could use a local workflow for publishing.

What are y'alls thoughts?

alexander-fenster:
Just a thought, instead of adding `GITHUB_TOKEN` directly, we can use kind of security through obscurity and commit encrypted script (containing the token), with encryption key added to CircleCI environment variables. Makes it just a little bit safer in case if CircleCI environment variables leak (but not 100% safe of course).

ofrobots:
Previously, a list of issues with mono-repos was compiled here: go/veneer-monorepo (internal link). It would be good to review the list and be wholly convinced that all the listed problems are solved or are no longer problems.

We already have a fair amount of tooling to deal with large number of repos. I am concerned that we are moving to a half-hearted-state where we have some `lerna` monorepos and some individual repos. IMO, this will cause a dilution in the tooling efforts – we need different kinds of tools to deal with the shortcoming of both `lerna` and individual repos.

kjin:
> For now, I think the process would be a human running `lerna publish`

Yep, from what I can tell `lerna publish` is not automatable unless you know that all packages should be released at the same time (which is slightly different than on the same version), and your commit messages are perfect.

JustinBeckwith:
Ok :) I created launchcals for each of the packages - once those are approved, I'll move this stuff out and submit a separate CL to remove the code from here.

PR #206, opened 2018-08-09
fix: method metadata can be a boolean

JustinBeckwith:


jkwlui (approved):
Thank you 🙏 Needed this for nodejs-storage as well!

PR #209, opened 2018-08-16
chore: remove dead code and packages

JustinBeckwith:
BREAKING CHANGE: This change removes the following methods from the `util` class:
- `util.replaceProjectIdToken`
- `util.extendGlobalConfig`
- `util.normalizeArguments`
- `util.createLimiter`
- `util.privatize`

These methods have either been replaced with smaller single purpose modules, or are no longer necessary. 

Side note: deleting code is fun, and feels really good.

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,6 +34,7 @@
     "Tim Swast <swast@google.com>"
   ],
   "scripts": {
+    "generate-scaffolding": "repo-tools generate all",
     "docs": "jsdoc -c .jsdoc.js",
     "cover": "nyc --reporter=lcov mocha build/test && nyc report",
     "test-no-cover": "mocha build/test",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -63,11 +61,9 @@
     "pify": "^4.0.0",
     "request": "^2.87.0",
     "retry-request": "^4.0.0",
-    "stream-events": "^1.0.4",
     "through2": "^2.0.3"
   },
   "devDependencies": {
```
fhinkel: IMHO this should be included so we can run `repo-tools generate all` every once in a while or we'll end up with the wrong list of contributors etc.

src/util.ts
```diff
@@ -714,152 +688,20 @@ export class Util {
...
+      reqOpts.json = replaceProjectIdToken(reqOpts.json, projectId);
     }
 
-    reqOpts.uri = util.replaceProjectIdToken(reqOpts.uri, projectId);
+    reqOpts.uri = replaceProjectIdToken(reqOpts.uri, projectId);
 
     return reqOpts;
   }
 
-  /**
```
fhinkel: :heart_eyes:

PR #210, opened 2018-08-17
feat: make HTTP dependency configurable

fhinkel:
`Service`, `ServiceObject`, and `util.makeRequest()` must be
initialized with an HTTP dependency, e.g., `request`.

The second parameter of `util.makeRequest`, a configuration
object, is mandatory.

Ultimately, this change will improve startup time for the
client libraries. Before this change, `nodejs-common` required
`request` as HTTP dependency, which is a very large dependency. By making the
HTTP dependency configurable, client libraries can use
smaller dependencies instead of `request`. Thus reducing startup time.

For an end-to-end integration, see https://github.com/googleapis/nodejs-translate/pull/98.

### DIFF
--- a/src/util.ts
+++ b/src/util.ts
@@ -31,14 +31,14 @@ import * as through from 'through2';
 
 import {Interceptor} from './service-object';
 
-const request = r.defaults({
+const requestDefaults = {
   timeout: 60000,
   gzip: true,
   forever: true,
   pool: {
     maxSockets: Infinity,
   },
-});
+};
 
 // tslint:disable-next-line:no-any
 export type ResponseBody = any;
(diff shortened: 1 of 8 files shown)


src/util.ts
```diff
@@ -31,15 +31,6 @@ import * as through from 'through2';
 
 import {Interceptor} from './service-object';
 
-const request = r.defaults({
```
JustinBeckwith: How are handling defaults moving forward?

fhinkel: They're part of teeny-request.

JustinBeckwith: So these specific defaults are part of teeny-request?  Or are we asking each individual library to set these?

PR #214, opened 2018-08-21
fix: increase timeout for install test

JustinBeckwith:
Fix all the nightlies!


kinwa91:
`npm install` takes more than 1 minute, so we might need a even higher timeout.

screen shot 2018-08-20 at 5 50 53 pm

PR #215, opened 2018-08-21
Remove old code & replace project ID token in multipart arrays.

stephenplusplus:
Fixes https://github.com/googleapis/nodejs-bigquery/issues/171

### DIFF
--- a/src/util.ts
+++ b/src/util.ts
@@ -158,15 +158,6 @@ export interface ParsedHttpResponseBody {
 }
 
 /**
- * Custom error type for missing project ID errors.
- */
-export class MissingProjectIdError extends Error {
-  message = `Sorry, we cannot connect to Cloud Services without a project
-    ID. You may specify one with an environment variable named
-    "GOOGLE_CLOUD_PROJECT".`.replace(/ +/g, ' ');
-}
-
-/**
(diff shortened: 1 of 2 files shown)


test/util.ts
```diff
@@ -1435,87 +1437,6 @@ describe('common/util', () => {
     });
   });
 
-  describe('projectId placeholder', () => {
```
stephenplusplus: These are tested in the `projectify` module.

test/util.ts
```diff
@@ -1386,8 +1360,10 @@ describe('common/util', () => {
       };
       const decoratedQs = {};
 
-      utilOverrides.replaceProjectIdToken = (qs: {}, projectId_: string) => {
```
stephenplusplus: This didn't actually have any effect, since we modularized out the projectify stuff.

src/util.ts
```diff
@@ -157,15 +157,6 @@ export interface ParsedHttpResponseBody {
   err?: Error;
 }
 
-/**
- * Custom error type for missing project ID errors.
- */
-export class MissingProjectIdError extends Error {
```
stephenplusplus: This is all in the projectify module now.

PR #235, opened 2018-09-14
Improve types for SO.getMetadata, setMetadata

kinwa91:
Downstream TypeScripted repos *might* have to update types.


src/service-object.ts
```diff
@@ -100,6 +100,10 @@ export interface GetConfig {
   autoCreate?: boolean;
 }
 
+export interface ResponseCallback {
```
JustinBeckwith: Could we also export this from index?

PR #239, opened 2018-09-18
Enhance typing of ServiceObject.prototype.get

carnesen:
Previously the type declarations for ServiceObject.prototype.get did not include signatures for the "promisified" version of the method. This PR adds suitable Promise-returning signatures. This also simplifies (`any`s) some of the `serviceObject.get` types in the unit tests, which are primarily designed to test run-time behavior rather than type design. 

- [x] Tests and linter pass

### DIFF
--- a/src/service-object.ts
+++ b/src/service-object.ts
@@ -106,8 +106,7 @@ export interface ResponseCallback {
 
 export type SetMetadataResponse = [r.Request];
 
-export type GetResponse =
-    [ServiceObject | null | undefined, r.Response | undefined];
+export type GetResponse = [ServiceObject, r.Response];
 
 /**
  * ServiceObject is a base class, meant to be inherited from by a "service
--- a/src/service-object.ts
+++ b/src/service-object.ts
@@ -106,6 +106,9 @@ export interface ResponseCallback {
 
 export type SetMetadataResponse = [r.Request];
 
+export type GetResponse =
+    [ServiceObject | null | undefined, r.Response | undefined];
+
 /**
  * ServiceObject is a base class, meant to be inherited from by a "service
  * object," like a BigQuery dataset or Storage bucket.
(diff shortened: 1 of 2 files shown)


test/service-object.ts
```diff
@@ -342,15 +339,10 @@ describe('ServiceObject', () => {
...
-          (configOrCallback: SO.GetConfig|SO.InstanceResponseCallback,
-           callback?: SO.InstanceResponseCallback) => {
-            callback = typeof configOrCallback === 'function' ?
-                configOrCallback :
-                callback;
-            callback!(error);
-          };
+      // tslint:disable-next-line:no-any
+      (serviceObject as any).get = (callback: SO.InstanceResponseCallback) => {
+        callback(error);
```
carnesen: It seemed to me that the previous code that mocked `serviceObject.get` was unnecessarily complex with the goal of being typed consistently with `serviceObject.prototype.get`. These "true-to-prototype" typings in the unit tests, however, failed to reveal that those typings were in fact incorrect. This PR therefore simplifies the mocked `serviceObject.get` definitions in the unit tests, preserving the runtime behavior but using the `any` escape valve for typings. I could be convinced (and already have some code toward this end on a branch) that instead we should abstract out all these `(serviceObject as any).get = ...` assignments into a lil helper function `mockGet` that assigns `serviceObject.get` in an appropriately-typed way.

src/service-object.ts
```diff
@@ -106,6 +106,9 @@ export interface ResponseCallback {
 
 export type SetMetadataResponse = [r.Response];
 
+export type GetResponse =
+    [ServiceObject | null | undefined, r.Response | undefined];
```
JustinBeckwith: In this case, `ServiceObject` shouldn't be null or undefined.  Same for the response I believe.  Just `[ServiceObject, r.Response]` will do fine :)

src/service-object.ts
```diff
@@ -276,10 +279,12 @@ class ServiceObject extends EventEmitter {
    * @param {object} callback.instance - The instance.
    * @param {object} callback.apiResponse - The full API response.
    */
-  get(config: GetConfig, callback?: InstanceResponseCallback): void;
+  get(): Promise<GetResponse>;
```
JustinBeckwith: We can collapse this line and line 248 by just marking `config` as optional.

JustinBeckwith (commented):
Thank you so much for the contribution!

src/service-object.ts
carnesen: Indeed! I saw the callback getting called with `null` serviceObject, but now that I take a closer look that's only for the error case.

JustinBeckwith (approved):

JustinBeckwith:
🥂

carnesen:
Yay! Thanks so much @JustinBeckwith for the code review!!

PR #241, opened 2018-09-19
Improve TypeScript types for async operations

### LINKED ISSUE #90: Add proper overloads for all promisified functions
Today we rely on `util.promisifyAll` to provide callback and promise style implementations.  While convenient at development time, it does lead to problems with TypeScript typing.  We should add proper overloads, and move towards a more async style for internal functions in the process.

JustinBeckwith:
Resolves #90


carnesen:
Does this close #90 ?

JustinBeckwith:
Yes it does!

PR #261, opened 2018-09-29
chore: Remove support for custom PromiseConstructor

### LINKED ISSUE #107: Drop Promise property
From #106  by @ofrobots 

> Not a comment for this PR, but in the next semver major, can we drop support from custom Promises? Note that async/await always use native promises, so the use-case for custom promises is growing thin, and might even be footgun as things behave differently from what people might expect.
> 
> Is there a strong reason to continue supporting this? People can always do global.Promise = that.thing;

carnesen:
BREAKING CHANGE: This removes support for custom localized `Promise` in `Service`, `ServiceObject`, and `Operation` in favor of consistently using `global.Promise`. Closes #107.

- [x] Tests and linter pass
- [?] Appropriate docs were updated (if necessary)

Regarding docs, I didn't see any in this repo specifically that needed updating. That doesn't mean they're not there!

### DIFF
--- a/src/operation.ts
+++ b/src/operation.ts
@@ -79,12 +79,11 @@ export class Operation extends ServiceObject {
    * @return {promise}
    */
   promise() {
-    return new this.Promise!
-        ((resolve: Function, reject: (err: Error) => void) => {
-          this.on('error', reject).on('complete', (metadata: {}) => {
-            resolve([metadata]);
-          });
-        });
+    return new Promise((resolve: Function, reject: (err: Error) => void) => {
+      this.on('error', reject).on('complete', (metadata: {}) => {
+        resolve([metadata]);
+      });
+    });
   }
 
   /**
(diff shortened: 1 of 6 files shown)


JustinBeckwith:
@googleapis/yoshi-nodejs if I understand this correctly - wouldn't this be a breaking change in every client library that takes a dependency?

carnesen:
@JustinBeckwith Yes, it's a breaking change. It sounded from https://github.com/googleapis/nodejs-common/issues/107 like the decision was to remove support for the "custom promise" feature. As a user, I always just use native Promises. As a contributor, I was just trying to close the issue. I'd be fine with closing this and the issue as "will not fix".

JustinBeckwith:
Apologies for the confusion here.  I don't think we're ready to make what amounts to a semver major change for each of our libs right now :(  I really appreciate the PR, but we're going to close out the issue for now.

Issue #263, opened 2018-10-01
Genericize types of ServiceObject prototype methods


carnesen:
We've been working on implementing proper TypeScript types for ServiceObject and its many subclasses. Last week we [released v0.8.0 of nodejs-dns](https://github.com/googleapis/nodejs-dns/releases/tag/v0.8.0) with "TypeScript support". I was kicking the tires on the new release and soon encountered a runtime error that is precisely the kind of thing that TypeScript is meant to avoid. [I filed an issue in nodejs-dns already](https://github.com/googleapis/nodejs-dns/issues/126), but ultimately I feel the solution is here in nodejs-common. Specifically I called `zone.create()`, which is perfectly acceptable [according to its type](https://github.com/googleapis/nodejs-common/blob/14a1ed14b9d2018b46d2e96592b7f61c3b19e27c/src/service-object.ts#L203), but was met with [a runtime error, "a zone dnsName is required"](https://github.com/googleapis/nodejs-dns/blob/d07673159609f2240394cf2a75a734f9ff546b08/src/index.ts#L231).

I think the best approach would be to use [TypeScript generics](https://www.typescriptlang.org/docs/handbook/generics.html) to parameterize the expected input type of ServiceObject.create. Specifically, rather than [typing ServiceObject's `config.createMethod` as `Function`](https://github.com/googleapis/nodejs-common/blob/14a1ed14b9d2018b46d2e96592b7f61c3b19e27c/src/service-object.ts#L65), we should type it as something like:
```TypeScript
interface CreateMethod<C> {
  (id: string, config: C): Promise<void>
}
```

carnesen:
Simplified example of how this might work: https://github.com/googleapis/nodejs-dns/issues/126#issuecomment-426024148

PR #265, opened 2018-10-11
fix: ServiceObject `get`, `create` should allow arbitrary config props

carnesen:
Before this change, for example, `dns.zone.get` does not allow a config argument `{ autoCreate: true, dnsName: 'carnesen.com' }`. It should allow that because the `ServiceObject.get` method just passes the config object through to the `create` method (sans the `autoCreate` property):
https://github.com/googleapis/nodejs-common/blob/0d5e5e2c371c3e37f324ebfbf9c95b836abd0f4e/src/service-object.ts#L333-L337

Note: A better approach long term would be to use some fancier typescript types as suggested in https://github.com/googleapis/nodejs-common/issues/263 . 

- [x] Tests and linter pass


src/service-object.ts
```diff
@@ -91,8 +91,6 @@ export interface Methods {
   [methodName: string]: {reqOpts?: r.CoreOptions}|boolean;
 }
 
-export interface CreateOptions {}
-
```
carnesen: An empty interfaces are a bad idea. Looking at it, my intuition tells me that a `CreateOptions` object must be empty. In fact, TypeScript does not do excess property checks on empty interfaces https://github.com/Microsoft/TypeScript/issues/14606 . If the intent here is to allow a _nearly_ arbitrary object as input to `ServiceObject.create`, better to type it explicitly as such as in this PR.

src/service-object.ts
```diff
@@ -102,10 +100,14 @@ export interface DeleteCallback {
...
+export interface CreateOptions {
+  // tslint:disable-next-line:no-any
+  [propertyName: string]: any;
+}
+
+export interface GetConfig extends CreateOptions {
+  // autoCreate:true creates the service object if it doesn't already exist
+  // in which case the rest of this config get passed into the "create" method
   autoCreate?: boolean;
 }
```
carnesen: The previous definition of `GetConfig` only allowed zero or one property (`autoCreate`) in the passed config object. In fact the "GetConfig" type should allow arbitrary additional parameters, which get passed into the the `ServiceObject`s 
 `createMethod`.

carnesen:
Ping ...

src/service-object.ts
JustinBeckwith: I don't think this is entirely right though.  Every TypeScript project that uses this library should be implementing their own GetOptions interface for each object type.  The `GetConfig` only really exists because I guess `autoCreate` is available on every service-object.  I don't think having `GetConfig` extend basically `any` is the right thing here 🤷‍♂️

carnesen:
This was meant as a stopgap to fix the type error googleapis/nodejs-dns#142 . Since there is some reluctance to accept this, I'll just work on the "real" solution using generics, which I now believe is within my grasp.

PR #268, opened 2018-10-15
fix: Do not retry streaming POST requests.

stephenplusplus:
Fixes https://github.com/googleapis/nodejs-storage/issues/27
Fixes https://github.com/googleapis/nodejs-storage/issues/312

Previously, we were retrying streaming POST requests, in places like `file.createWriteStream()`. That's not actually possible, since once the data has left the user's stream, and we pass it to the upstream API, it's effectively gone to free up memory. So if suddenly the API decides it's upset, we can't rewind the state of the transfer.

### DIFF
--- a/system-test/fixtures/kitchen/src/index.ts
+++ b/system-test/fixtures/kitchen/src/index.ts
@@ -4,7 +4,7 @@ import {GoogleAuthOptions, Operation,Service, ServiceConfig, ServiceOptions,
   ServiceObjectConfig, StreamRequestOptions, Abortable, AbortableDuplex,
   ApiError, util} from '@google-cloud/common';
 
-util.makeRequest({url: 'test'}, {}, (err, body, res) => {
+util.makeRequest({uri: 'test'}, {}, (err, body, res) => {
   console.log(err);
 });
 
--- a/test/util.ts
+++ b/test/util.ts
@@ -403,9 +403,10 @@ describe('common/util', () => {
       const metadata = {a: 'b', c: 'd'} as any;
       util.makeWritableStream(dup, {
         metadata,
-        makeAuthenticatedRequest(request) {
+        makeAuthenticatedRequest(request: DecorateRequestOptions) {
           assert.equal(request.method, 'POST');
           assert.equal(request.qs.uploadType, 'multipart');
+          assert.strictEqual(request.maxRetries, 0);
 
           assert.strictEqual(Array.isArray(request.multipart), true);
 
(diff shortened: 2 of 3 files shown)


test/util.ts
```diff
@@ -1092,33 +1093,6 @@ describe('common/util', () => {
       };
     }
 
-    describe('callback mode', () => {
-      it('should pass the default options to retryRequest', (done) => {
```
stephenplusplus: These tests are already in the `describe('callback mode` that exists below the `describe('stream mode`.

test/util.ts
```diff
@@ -1092,33 +1093,6 @@ describe('common/util', () => {
...
-        retryRequestOverride = testNoRetryRequestConfig(done);
-        util.makeRequest(reqOpts, noRetryRequestConfig, () => {});
-      });
-
-      it('should override number of retries to retryRequest', (done) => {
-        retryRequestOverride = testCustomRetryRequestConfig(done);
-        util.makeRequest(reqOpts, customRetryRequestConfig, () => {});
-      });
-
-      it('should return the instance of retryRequest', () => {
```
stephenplusplus: This was the one test that wasn't duplicated, so I just moved it down with the rest.

stephenplusplus:
@JustinBeckwith I was getting this locally, but was hoping it was just a quirk of my environment:

```
1) should be able to use the d.ts:
     Error: Spawn failed with an exit code of 1
      at ChildProcess.<anonymous> (build/test/install.js:15:25476)
      at maybeClose (internal/child_process.js:925:16)
      at Process.ChildProcess._handle.onexit (internal/child_process.js:209:5)
```

Do you know this one? Any ideas where to look?

JustinBeckwith:
Ah, I think I see the problem.  `DecorateRequestOptions` is extending `r.OptionsWithUri`, while `r.Options` is `r.OptionsWithUri | r.OptionsWithUrl`.  The fix here is probably to have `DecorageRequestOptions` extend `r.Options`.

system-test/fixtures/kitchen/src/index.ts
```diff
@@ -4,7 +4,7 @@ import {GoogleAuthOptions, Operation,Service, ServiceConfig, ServiceOptions,
   ServiceObjectConfig, StreamRequestOptions, Abortable, AbortableDuplex,
   ApiError, util} from '@google-cloud/common';
 
-util.makeRequest({url: 'test'}, {}, (err, body, res) => {
+util.makeRequest({uri: 'test'}, {}, (err, body, res) => {
```
stephenplusplus: @JustinBeckwith is this cool? This was the sole fix for all of that monospace sadness.

JustinBeckwith: s'all good

micahwedemeyer:
Any idea when you'll release a new version of the `cloud-storage` package? I'm eagerly awaiting a chance to update to the latest and see if it fixes the dreaded ECONNRESET bugs.

stephenplusplus:
`@google-cloud/storage` v2.2.0 was released, which includes this fix.

PR #270, opened 2018-10-16
Remove 'is' dependency

### LINKED ISSUE #163: Migrate to @sindresorhus/is
We should use `@sindresorhus/is` instead of just `is`, because it's more consistent.  PR feedback from https://github.com/googleapis/nodejs-common/pull/162

zbjornson:
All of these replacements are synonymous with `is`'s implementations -- all of `is`'s tests pass using these versions.

Closes #163

- [x] Tests and linter pass

### DIFF
--- a/package.json
+++ b/package.json
@@ -69,7 +69,6 @@
     "ent": "^2.2.0",
     "extend": "^3.0.1",
     "google-auth-library": "^2.0.0",
-    "is": "^3.2.1",
     "pify": "^4.0.0",
     "retry-request": "^4.0.0",
     "through2": "^2.0.3"
@@ -79,7 +78,6 @@
     "@types/arrify": "^1.0.4",
     "@types/ent": "^2.2.1",
     "@types/extend": "^3.0.0",
-    "@types/is": "0.0.20",
     "@types/mocha": "^5.2.1",
     "@types/mv": "^2.1.0",
     "@types/ncp": "^2.0.1",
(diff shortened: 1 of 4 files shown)


zbjornson:
Knowing how to check the type of simple, built-in JS values I think is fundamental knowledge. TS is smart enough to know that [`typeof`](https://www.typescriptlang.org/docs/handbook/advanced-types.html#typeof-type-guards), [`instanceof`](https://www.typescriptlang.org/docs/handbook/advanced-types.html#instanceof-type-guards) and `Array.isArray` narrow the type. I like avoiding dependencies for trivial tasks to make installs fast, keep node_modules small (in this case saving 49 KB), speedup node.js startup, reduce time maintainers have to spend bumping dependencies, and avoid potential upstream bugs and vulnerabilities. (The original `is` library, which is also tiny and popular, has a runtime strict-mode violation that is only not encountered because of another bug in that module, for example.)

That said, obviously not my final decision. 😉

JustinBeckwith:
@carnesen I see where you're coming from, but I like this for a few reasons:
1. The TypeScript compiler can recognize standard typeof guards when evaluating types
2. It's one less dependency in the chain for a very, very foundational library

This is cool to me 👍

zbjornson:
Thanks, both fixed. 🙂 

(@carnesen didn't mean my comment to be rude, sorry if it came off strong.)

carnesen:
@zbjornson No worries at all. I don't mince words either! This is basically just a "leftpad" debate, and I was voicing my vote in favor of the "module" side.

PR #303, opened 2018-11-15
feat: release @google-cloud/common v0.26.2

kinwa91:
Making this a patch release so that the nodejs-projectify fix would go out to common's dependents.


kinwa91:
Release is at https://github.com/googleapis/nodejs-common/releases/tag/v0.26.2

PR #315, opened 2018-12-03
feat: add maybeOptionsOrCallback util method

JustinBeckwith:
We use this pattern all over the place, and it would be good to have a single tested implementation :)


JustinBeckwith:
cc @ofrobots

PR #334, opened 2018-12-14
fix: flip order of request callback params

JustinBeckwith:
So here I am, trying to figure out what in God's green earth is happening in the compute tests.  I'm looking at an overridden `request` function that assumes a callback like `(err, response, body) => {}`, and I'm all like "what?  this is so weird.  That's not what the TypeScript types say the function signature should be."  Then I sit and think...  but wait, that's the actual order of params from the request module.  What in the hell?

Way back when we (fine, I) converted this module over to TypeScript, I *reversed the order of the request callback parameters*.  The fact that `body` is of type `any`, meant that technically anything was accepted in that parameter.  I feel silly.  I feel bad.  But anyway...  this ought to fix things.

### DIFF
--- a/src/service-object.ts
+++ b/src/service-object.ts
@@ -25,17 +25,16 @@ import * as extend from 'extend';
 import * as r from 'request';  // Only needed for type declarations.
 
 import {StreamRequestOptions} from '.';
-import {ApiError, BodyResponseCallback, DecorateRequestOptions, util} from './util';
+import {ApiError, DecorateRequestOptions, util} from './util';
 
 export type CreateOptions = {};
-export type RequestResponse = [Metadata, r.Response];
+export type RequestResponse = [r.Response, Metadata];
 
 export interface ServiceObjectParent {
   // tslint:disable-next-line:variable-name
   Promise?: PromiseConstructor;
   requestStream(reqOpts: DecorateRequestOptions): r.Request;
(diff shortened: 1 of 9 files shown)


src/service-object.ts
```diff
@@ -44,8 +43,11 @@ export interface Interceptor {
 
 // tslint:disable-next-line:no-any
 export type Metadata = any;
-export type MetadataResponse = [Metadata, r.Response];
-export type MetadataCallback =
+export type GetMetadataResponse = [Metadata, r.Response];
```
alexander-fenster: So just to confirm, `RequestResponse` is `[r.Response, Metadata]` but `GetMetadataResponse` is `[Metadata, r.Response]`? :)

alexander-fenster: Well, it actually makes sense after some thinking, since metadata is secondary for `RequestResponse` and is primary for `GetMetadataResponse`, but still... :)

alexander-fenster:
I just wonder if it's a breaking change or not. I see you're swapping parameters in some callbacks which looks pretty scary for me. Can we at least run system tests for all or some modules that depend on `common` (by making them depend on `"@google-cloud/common": "JustinBeckwith/nodejs-common#i-feel-bad")?

JustinBeckwith:
Superseded by #335

PR #335, opened 2018-12-14
fix: reverse order of request callback params

JustinBeckwith:
Hilariously the order of types for the request callback have been wrong for some time.  This change just flips the order of the types, and should have no actual runtime impact.


kinwa91:
note to self: requires a minor release, breaks downstream TypeScript compilation.

PR #338, opened 2018-12-15
fix: Revert "fix: reverse order of request callback params (#335)"

JustinBeckwith:
This reverts commit a732f8dfbcf8aa68c8d04ef89890d166506baa20.


JustinBeckwith:
TBR = @kinwa91 

My original PR to do this was a very bad no good mistake.

PR #344, opened 2018-12-23
fix: inherit requestModule from parent

### LINKED ISSUE #222: ServiceObject should inherit parent's requestModule
Original discussion: https://github.com/googleapis/nodejs-bigquery/pull/175/files#r212806493

JustinBeckwith:
Resolves #222

### DIFF
--- a/src/service-object.ts
+++ b/src/service-object.ts
@@ -33,6 +33,7 @@ export type RequestResponse = [Metadata, r.Response];
 export interface ServiceObjectParent {
   // tslint:disable-next-line:variable-name
   Promise?: PromiseConstructor;
+  requestModule?: typeof r;
   requestStream(reqOpts: DecorateRequestOptions): r.Request;
   request(reqOpts: DecorateRequestOptions, callback: BodyResponseCallback):
       void;
@@ -83,7 +84,7 @@ export interface ServiceObjectConfig {
   /**
    * Dependency for HTTP calls.
    */
-  requestModule: typeof r;
+  requestModule?: typeof r;
 }
 
 export interface Methods {
@@ -171,7 +172,8 @@ class ServiceObject<T = any> extends EventEmitter {
     this.methods = config.methods || {};
     this.interceptors = [];
     this.Promise = this.parent ? this.parent.Promise : undefined;
-    this.requestModule = config.requestModule;
+    this.requestModule =
+        (config.requestModule || (this.parent && this.parent.requestModule))!;
 
     if (config.methods) {
       Object.getOwnPropertyNames(ServiceObject.prototype)
--- a/test/service-object.ts
+++ b/test/service-object.ts
@@ -44,10 +44,11 @@ describe('ServiceObject', () => {
 
   const CONFIG = {
     baseUrl: 'base-url',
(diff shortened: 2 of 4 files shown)


JustinBeckwith:
*gentle ping* @stephenplusplus

src/service-object.ts
```diff
@@ -171,7 +172,13 @@ class ServiceObject<T = any> extends EventEmitter {
     this.methods = config.methods || {};
     this.interceptors = [];
     this.Promise = this.parent ? this.parent.Promise : undefined;
-    this.requestModule = config.requestModule;
+    const requestModule =
+        config.requestModule || (this.parent && this.parent.requestModule);
+    if (!requestModule) {
+      throw new Error(
```
stephenplusplus: Historically, I've avoided throwing dev-stuff like this. Through other tests, we should be able to prevent this from making it to the end user (system tests, definitely).

test/service-object.ts
```diff
@@ -114,13 +115,26 @@ describe('ServiceObject', () => {
...
       assert.strictEqual(serviceObject.Promise, FakePromise);
     });
+
+    it('should inherit the parents requestModule', () => {
+      const serviceObject = new ServiceObject(CONFIG);
+      assert.strictEqual(
+          serviceObject.requestModule, CONFIG.parent.requestModule);
+    });
+
+    it('should throw if a requestModule is not passed to the constructor',
```
stephenplusplus: Thank goodness for auto-formatters, 'cause imagine trying to explain the rules of the next several lines to a contributor 🤣

JustinBeckwith: software is terrible.

src/service-object.ts
JustinBeckwith: Hmm, so I kinda want to poke on this one a little :) I pushed this throw forward because all we're doing is deferring a thrown exception.  Any time the code goes down a path of making an http request, this is just going to fail at a later time.  Why wouldn't we want to push the failure forward to construction?

stephenplusplus: This is an us-failure, which I don't think the user should error handle. That's for us (through testing)! API errors or user errors are for the end user to worry about. @callmehiphop any thoughts?

stephenplusplus: I had a side chat with @callmehiphop, and he brought up this error would be valuable to users who might have `common` in their own, non-official libraries. That's very true, because then this becomes an error that is intended for the "user" of common. But we don't actually support that, even though it looks like we have been used by other libraries, in some form: https://www.npmjs.com/browse/depended/@google-cloud/common

So, my not-*that*-strong opinion on this matter is that since we are a "private" library, intended for use within our own libraries, we shouldn't have code that looks like we support something that we don't actually support. We don't have docs that we expose for this library (I don't think?), so I think it's fair to say that this is just our internal tool we use for our own purposes.

Operating from that premise, I think this is an error that a user of `@google-cloud/storage` should never see. Our system tests in Storage wouldn't pass if we didn't have `requestModule`, so I feel safe relying on those tests instead of this one.

JustinBeckwith:
👋 @stephenplusplus got it - PTAL?

PR #349, opened 2019-01-18
feat: allow options to ServiceObject methods

stephenplusplus:
#### To Dos
- [x] Tests

This allows `options` to be passed to several methods on ServiceObject:

- [x] delete
- [x] exists
- [x] get
- [x] getMetadata
- [x] setMetadata

For Storage, in order to support the `userProject` object, we ended up duplicating many of these methods, so that we could pass extra parameters along with the request as query string parameters. With this change, we will just support that here, where it probably belongs.

Sending this in early for feedback before I carry on with the other methods and tests.

### DIFF
--- a/src/service-object.ts
+++ b/src/service-object.ts
@@ -103,6 +103,7 @@ export interface CreateCallback<T> {
   (err: ApiError|null, instance?: T|null, ...args: any[]): void;
 }
 
+export type DeleteOptions = object;
 export interface DeleteCallback {
   (err: Error|null, apiResponse?: r.Response): void;
 }
@@ -249,17 +250,21 @@ class ServiceObject<T = any> extends EventEmitter {
    * @param {?error} callback.err - An error returned while making this request.
    * @param {object} callback.apiResponse - The full API response.
    */
-  delete(): Promise<[r.Response]>;
+  delete(options?: DeleteOptions): Promise<[r.Response]>;
+  delete(options: DeleteOptions, callback: DeleteCallback): void;
   delete(callback: DeleteCallback): void;
-  delete(callback?: DeleteCallback): Promise<[r.Response]>|void {
+  delete(optionsOrCallback: DeleteOptions|DeleteCallback, cb?: DeleteCallback): Promise<[r.Response]>|void {
+    const [options, callback] = util.maybeOptionsOrCallback<DeleteOptions, DeleteCallback>(
+        optionsOrCallback, cb);
+
     const methodConfig =
         (typeof this.methods.delete === 'object' && this.methods.delete) || {};
-    callback = callback || util.noop;
 
     const reqOpts = extend(
         {
           method: 'DELETE',
           uri: '',
+          qs: options,
         },
         methodConfig.reqOpts);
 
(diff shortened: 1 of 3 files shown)


src/service-object.ts
```diff
@@ -42,8 +42,13 @@ export interface Interceptor {
   request(opts: r.Options): DecorateRequestOptions;
 }
 
+// tslint:disable-next-line:no-any
+export type GetMetadataOptions = any;
```
callmehiphop: I think maybe it would be better to use `object` than `any` here. I think it is safe to assume that we'll only ever allow objects to be passed in (as opposed to strings, numbers, etc.).

src/service-object.ts
```diff
@@ -42,8 +42,13 @@ export interface Interceptor {
...
+
 // tslint:disable-next-line:no-any
 export type Metadata = any;
+// tslint:disable-next-line:no-any
+export type SetMetadataOptions = any;
```
callmehiphop: See comments for `GetMetadataOptions`.

src/service-object.ts
```diff
@@ -360,13 +365,28 @@ class ServiceObject<T = any> extends EventEmitter {
...
-  getMetadata(): Promise<MetadataResponse>;
+  getMetadata(options?: GetMetadataOptions): Promise<MetadataResponse>;
   getMetadata(callback: MetadataCallback): void;
-  getMetadata(callback?: MetadataCallback): Promise<MetadataResponse>|void {
+  getMetadata(options: GetMetadataOptions, callback: MetadataCallback): void;
+  getMetadata(
+      optionsOrCallback?: GetMetadataOptions|MetadataCallback,
+      callback?: MetadataCallback): Promise<MetadataResponse>|void {
+    let options: GetMetadataOptions = {};
+    if (typeof optionsOrCallback === 'function') {
```
callmehiphop: Consider using [maybeOptionsOrCallback](https://github.com/googleapis/nodejs-common/blob/master/src/util.ts#L756) to simplify this logic!

src/service-object.ts
```diff
@@ -382,14 +402,25 @@ class ServiceObject<T = any> extends EventEmitter {
...
+      metadata: Metadata, options: SetMetadataOptions,
+      callback: MetadataCallback): void;
   setMetadata(metadata: Metadata, callback: MetadataCallback): void;
-  setMetadata(metadata: Metadata, callback?: MetadataCallback):
-      Promise<SetMetadataResponse>|void {
+  setMetadata(
+      metadata: Metadata,
+      optionsOrCallback?: SetMetadataOptions|MetadataCallback,
+      callback?: MetadataCallback): Promise<SetMetadataResponse>|void {
+    const options =
```
callmehiphop: Consider using [maybeOptionsOrCallback](https://github.com/googleapis/nodejs-common/blob/master/src/util.ts#L756) to simplify this logic!

src/service-object.ts
```diff
@@ -42,8 +42,13 @@ export interface Interceptor {
   request(opts: r.Options): DecorateRequestOptions;
 }
 
+// tslint:disable-next-line:no-any
```
callmehiphop: I think you can get rid of this now?

stephenplusplus: 😊

stephenplusplus:
@JustinBeckwith @callmehiphop -- tests added. PTAfinalL. It would be great to get this in before the next release (https://github.com/googleapis/nodejs-common/pull/350).

PR #365, opened 2019-02-06
fix: remove timeout rule from streaming uploads

stephenplusplus:
Fixes https://github.com/googleapis/nodejs-storage/issues/27


stephenplusplus:
I believe so. Removing the `timeout` should mean that node-fetch will just let the request play out. Alternatively, we could set a larger timeout value.

PR #378, opened 2019-02-26
fix(autoCreate): leave user options intact

callmehiphop:
Fixes https://github.com/googleapis/nodejs-bigquery/issues/362

- [ ] Tests and linter pass

### DIFF
--- a/src/service-object.ts
+++ b/src/service-object.ts
@@ -324,8 +324,9 @@ class ServiceObject<T = any> extends EventEmitter {
       cb?: InstanceResponseCallback<T>): Promise<GetResponse<T>>|void {
     const self = this;
 
-    const [options, callback] = util.maybeOptionsOrCallback<
+    const [opts, callback] = util.maybeOptionsOrCallback<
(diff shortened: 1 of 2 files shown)


src/service-object.ts
```diff
@@ -324,8 +324,9 @@ class ServiceObject<T = any> extends EventEmitter {
       cb?: InstanceResponseCallback<T>): Promise<GetResponse<T>>|void {
     const self = this;
 
-    const [options, callback] = util.maybeOptionsOrCallback<
+    const [opts, callback] = util.maybeOptionsOrCallback<
         GetOrCreateOptions, InstanceResponseCallback<T>>(optionsOrCallback, cb);
+    const options = Object.assign({}, opts);
```
stephenplusplus: I think we need this in all of the places in the file where autoCreate is used.

stephenplusplus: Which is just here, obviously. Did everyone see this in their email? I'm just gonna delete this comment and nobody will know.

Issue #392, opened 2019-03-27
@types included in dependencies


niksoper:
#### Environment details

  - OS: Mac Mojave
  - Node.js version: `8.9.1` | `8.15.0` | `10.14.0`
  - npm version: `5.5.1` | `6.4.1`
  - `@google-cloud/common` version: `0.31.1`

#### Steps to reproduce

  1. Create empty npm package:`npm init`
  2. Add @google-cloud/common as a dependency: `yarn add @google-cloud/common@0.31.1`

We now have this `package.json`:
```
{
  "name": "temp",
  "version": "1.0.0",
  "description": "",
  "main": "index.js",
  "scripts": {
    "test": "echo \"Error: no test specified\" && exit 1"
  },
  "author": "",
  "license": "ISC",
  "dependencies": {
    "@google-cloud/common": "0.31.1"
  }
}
```

  4. List all TypeScript typings from dependencies: `ls node_modules/@types`:
```
caseless     duplexify    form-data    node         request      tough-cookie
```

  3. Check why we have these:
```
yarn why @types/request

yarn why v1.13.0
[1/4] 🤔  Why do we have the module "@types/request"...?
[2/4] 🚚  Initialising dependency graph...
[3/4] 🔍  Finding dependency...
[4/4] 🚡  Calculating file sizes...
=> Found "@types/request@2.48.1"
info Reasons this module exists
   - "@google-cloud#common" depends on it
   - Hoisted from "@google-cloud#common#@types#request"
info Disk size without dependencies: "28KB"
info Disk size with unique dependencies: "796KB"
info Disk size with transitive dependencies: "796KB"
info Number of shared dependencies: 4
✨  Done in 0.11s.
```
```
yarn why @types/duplexify

yarn why v1.13.0
[1/4] 🤔  Why do we have the module "@types/duplexify"...?
[2/4] 🚚  Initialising dependency graph...
[3/4] 🔍  Finding dependency...
[4/4] 🚡  Calculating file sizes...
=> Found "@types/duplexify@3.6.0"
info Reasons this module exists
   - "@google-cloud#common" depends on it
   - Hoisted from "@google-cloud#common#@types#duplexify"
info Disk size without dependencies: "16KB"
info Disk size with unique dependencies: "732KB"
info Disk size with transitive dependencies: "732KB"
info Number of shared dependencies: 1
✨  Done in 0.11s.
```

[...]

PR #394, opened 2019-03-27
refactor: remove configurable request

JustinBeckwith:
**BREAKING CHANGE**: This PR removes the ability to configure a custom implementation of the `Request` module. This was necessary when we were migrating from `request` to `teeny-request`, but that migration is now complete.  All interfaces at accepted a custom implementation of `request` will no longer accept one. `teeny-request` is now just included in the box.

This change is being made to simplify the `request` interface touch points.

### DIFF
--- a/src/service-object.ts
+++ b/src/service-object.ts
@@ -32,7 +32,6 @@ export type RequestResponse = [Metadata, r.Response];
 export interface ServiceObjectParent {
   // tslint:disable-next-line:variable-name
   Promise?: PromiseConstructor;
-  requestModule?: typeof r;
   requestStream(reqOpts: DecorateRequestOptions): r.Request;
   request(reqOpts: DecorateRequestOptions, callback: BodyResponseCallback):
       void;
@@ -82,11 +81,6 @@ export interface ServiceObjectConfig {
    * object is Bucket.
    */
   parent: ServiceObjectParent;
-
-  /**
-   * Dependency for HTTP calls.
-   */
-  requestModule?: typeof r;
 }
 
 export interface Methods {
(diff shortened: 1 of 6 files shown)


JustinBeckwith:
FYI @ofrobots, this is in pursuit of #392

src/service-object.ts
```diff
@@ -178,8 +171,6 @@ class ServiceObject<T = any> extends EventEmitter {
     this.methods = config.methods || {};
     this.interceptors = [];
     this.Promise = this.parent ? this.parent.Promise : undefined;
-    this.requestModule =
```
ofrobots: It may be nice to throw if `config.requestModule` is set.

JustinBeckwith: I seem to remember @stephenplusplus smacking my hands last time I did something like that in this exact spot :)

stephenplusplus: Haha. Well, throwing developer errors hasn't been something we have done before. We only throw errors that make sense to the user, and leave it to our various testing mechanisms to cover how we use our own libraries.

ofrobots: This is a different use-case though. We are detecting that the user didn't read the changelog and is passing a config flag that we are going to silently ignore. It would be nice to tell them that there was a breaking change. I don't have objections to the general disposition about throwing errors in these libraries; but this is different.

stephenplusplus: The users are us, though.

JustinBeckwith: I am wildly irresponsible and really rely on TypeScript types and aggressive exceptions to save me 🤷‍♂️

bcoe: I've seen (and like) the pattern of asserting that arguments/config are correct at the start of a function, so that a client fails as early as possible when interacting with the library (vs., having undefined behavior that bubbles as an error like can't access property of undefined).

Having said this, given we're calling this a breaking change and, like @stephenplusplus says, this is an interface probably mainly used by ourselves, I'd argue against the throw (which we'll then wonder about why the heck it's in the codebase for the next 5 years).

PR #403, opened 2019-04-04
fix: teeny-request should be dependency

bcoe:
Fixes issues seen here: https://github.com/googleapis/cloud-debug-nodejs/pull/667

teeny-request is required in util.js, but is only currently present as a dev-dependency.

- [x] Tests and linter pass


kjin:
Can we have a new release with this fix? :)

Issue #426, opened 2019-05-21
Improve ServiceObject types


callmehiphop:
👋 

I'm implementing a new feature that uses the `ServiceObject` class and have found myself getting frustrated with some of the typings. I have a few thoughts on how we could improve the types here and would love to get some yoshisaur opinions.

## Use `this` instead of a generic type variable

The [`ServiceObject` accepts a generic type variable](https://github.com/googleapis/nodejs-common/blob/master/src/service-object.ts#L146) that is used to pass around a reference to whatever the type that extends it is. TypeScript actually supports a `this` type, which seems more appropriate in this case.

## Use a generic type variable for metadata

Every `ServiceObject` instance has a `metadata` property that is just the JSON representation of a response body. Right now it is set as [an alias to `any`](https://github.com/googleapis/nodejs-common/blob/master/src/service-object.ts#L55), which gives our users no type safety when accessing members on it. Using a generic here makes sense to me.

## Allow for a generic type variable in the `request()` function

I think it would be nice if we didn't have to explicitly set types in request callbacks, I think adding a generic type variable to the [`request` method](https://github.com/googleapis/nodejs-common/blob/master/src/service-object.ts#L530) would be a great fix here.

JustinBeckwith:
Initial thoughts:
- *Use this instead of a generic type variable* - yep, go for it!
- *Use a generic type variable for metadata* - This should be optional so it's not a breaking change, but I'm otherwise fine with it.  Would be good to see an example.
- *Allow for a generic type variable in the request() function* - same deal as before - this should default to `any`, but let you override it.  We do this with `gaxios`, and it's great 🙃

sofisl:
Closing for now, feel free to reopen if this FR is still needed.

PR #447, opened 2019-07-08
refactor!: use `this` types in service-object

JustinBeckwith:
A hat tip towards #426.  Soooooo not ready yet.

### DIFF
--- a/src/service-object.ts
+++ b/src/service-object.ts
@@ -539,9 +539,7 @@ class ServiceObject extends EventEmitter {
    * @param {string} reqOpts.uri - A URI relative to the baseUrl.
    * @param {function} callback - The callback function passed to `request`.
    */
-  request(
-    reqOpts: DecorateRequestOptions
-  ): Promise<RequestResponse>;
+  request(reqOpts: DecorateRequestOptions): Promise<RequestResponse>;
   request(
     reqOpts: DecorateRequestOptions,
     callback: BodyResponseCallback
--- a/src/service-object.ts
+++ b/src/service-object.ts
@@ -33,7 +33,7 @@ import {
   util,
 } from './util';
 
-export type RequestResponse<T = ResponseBody> = [Metadata, r.Response];
+export type RequestResponse = [Metadata, r.Response];
 
 export interface ServiceObjectParent {
   // tslint:disable-next-line:variable-name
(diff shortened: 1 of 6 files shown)


src/service-object.ts
```diff
@@ -525,7 +520,9 @@ class ServiceObject<T = any> extends EventEmitter {
    * @param {string} reqOpts.uri - A URI relative to the baseUrl.
    * @param {function} callback - The callback function passed to `request`.
    */
-  request(reqOpts: DecorateRequestOptions): Promise<RequestResponse>;
+  request<T = ResponseBody>(
+    reqOpts: DecorateRequestOptions
+  ): Promise<RequestResponse>;
```
callmehiphop: Should this be `Promise<RequestResponse<T>>`?

jkwlui: I think this is half-finished - I'm going to revert this, and open another PR to genericize the request methods.

src/service-object.ts
```diff
@@ -99,7 +99,7 @@ export interface Methods {
 }
 
 export interface InstanceResponseCallback<T> {
-  (err: ApiError | null, instance?: T | null, apiResponse?: r.Response): void;
+  (err: ApiError | null, instance?: T | null, apiResponse?: Metadata): void;
```
jkwlui: We need to decide what the correct return type for `get` is.

src/service-object.ts
jkwlui: ..as part 3 of #426

bcoe:
@callmehiphop @JustinBeckwith what are the odds of dusting this off at some point? wonder who would be best to hand it off to ... @callmehiphop you seem to have a bit more context than me.

JustinBeckwith:
The work that really needs to happen here is vetting the change with a few dependents, like nodejs-storage, nodejs-dns, etc.  I seem to remember not being about to make the types work when I tried it.

callmehiphop:
@JustinBeckwith @bcoe any other clients we'd like to test against besides storage and dns? I'm happy to take a crack at this

JustinBeckwith:
`nodejs-storage` and `cloud-profiler-nodejs` are great candidates.  They are the ones I always seem to break 😆 

In all honesty - I seem to remember coming to the conclusion this approach wouldn't work after trying to update nodejs-storage. Please feel free to correct me though!

JustinBeckwith:
I just couldn't make this work, and this PR ain't aging like a fine wine.  Closing this out for now.

PR #449, opened 2019-07-08
Includes the response in the error message when the response body can't be parsed as JSON.

### LINKED ISSUE #446: parseHttpRespBody could provide a more helpful error message when the response can't be parsed as JSON
**Is your feature request related to a problem? Please describe.**

I see the following error message in my code very occasionally. I think it's coming from [here](https://github.com/googleapis/nodejs-common/blob/8e73d8cc9fc96a19690cfa947192c6b5aaee048b/src/util.ts#L399-L428).

```
 Error: Cannot parse JSON response                                                                                                                                                                                                            +
     at Util.parseHttpRespBody (/app/node_modules/@google-cloud/common/build/src/util.js:186:42)                                                                                                                                              +
     at Util.handleResp (/app/node_modules/@google-cloud/common/build/src/util.js:135:117)                                                                                                                                                    +
     at /app/node_modules/@google-cloud/common/build/src/util.js:242:26                                                                                                                                                                       +
     at /app/node_modules/teeny-request/build/src/index.js:156:17                                                                                                                                                                             +
     at processTicksAndRejections (internal/process/task_queues.js:85:5)
```

 **Describe the solution you'd like**

I'd like the response that couldn't be JSON-parsed to be surfaced, either in the error message, or in logs. I think I'd prefer to see it in the error message. It'd make troubleshooting easier, and it would allow me to write error-dependent retries.

I know that Google APIs sometimes return generic HTML errors, but it's hard to tell if that's what's happening here.

If it sounds okay to add the error message, I'd be happy to make a PR. Let me know!

Tyler-Murphy:
Fixes #446.

- [x] Tests and linter pass

before:

File               |  % Stmts | % Branch |  % Funcs |  % Lines | Uncovered Line #s |
-------------------|----------|----------|----------|----------|-------------------|
All files          |    99.24 |    97.86 |      100 |    99.23 |                   |
 index.ts          |      100 |      100 |      100 |      100 |                   |
 operation.ts      |      100 |      100 |      100 |      100 |                   |
 service-object.ts |      100 |    97.92 |      100 |      100 |               184 |
 service.ts        |    98.15 |    95.24 |      100 |    98.11 |               232 |
 util.ts           |    98.97 |     97.9 |      100 |    98.96 |           367,418 |

after:

File               |  % Stmts | % Branch |  % Funcs |  % Lines | Uncovered Line #s |
-------------------|----------|----------|----------|----------|-------------------|
All files          |    99.49 |    97.86 |      100 |    99.49 |                   |
 index.ts          |      100 |      100 |      100 |      100 |                   |
 operation.ts      |      100 |      100 |      100 |      100 |                   |
 service-object.ts |      100 |    97.92 |      100 |      100 |               184 |
 service.ts        |    98.15 |    95.24 |      100 |    98.11 |               232 |
 util.ts           |    99.49 |     97.9 |      100 |    99.48 |               367 |

  - I didn't see any that were relevant, but I might have missed them

### DIFF
--- a/src/util.ts
+++ b/src/util.ts
@@ -367,6 +367,10 @@ export class Util {
       parsedResp.resp.body = parsedResp.body;
     }
 
+    if (parsedResp.err && resp) {
+      parsedResp.err.response = resp;
+    }
+
     callback(parsedResp.err, parsedResp.body, parsedResp.resp);
   }
 
--- a/src/util.ts
+++ b/src/util.ts
@@ -300,7 +300,7 @@ export class PartialFailureError extends Error {
 }
 
 export interface BodyResponseCallback {
-  (err: Error | null, body?: ResponseBody, res?: r.Response): void;
+  (err: Error | ApiError | null, body?: ResponseBody, res?: r.Response): void;
 }
 
 export interface MakeRequestConfig {
--- a/src/util.ts
+++ b/src/util.ts
@@ -404,7 +404,7 @@ export class Util {
    * @param {?error} parsedHttpRespMessage.err - An error detected.
    * @param {object} parsedHttpRespMessage.body - The original body value provided
    *     will try to be JSON.parse'd. If it's successful, the parsed value will
-   * be returned here, otherwise the original value.
+   * be returned here, otherwise the original value and an error will be returned.
    */
   parseHttpRespBody(body: ResponseBody) {
     const parsedHttpRespBody: ParsedHttpResponseBody = {
@@ -415,7 +415,9 @@ export class Util {
       try {
         parsedHttpRespBody.body = JSON.parse(body);
       } catch (err) {
-        parsedHttpRespBody.err = new ApiError('Cannot parse JSON response');
+        parsedHttpRespBody.err = new ApiError(
+          `Cannot parse response as JSON: ${body}`
+        );
       }
     }
 
--- a/test/util.ts
+++ b/test/util.ts
@@ -377,6 +377,20 @@ describe('common/util', () => {
       stub('parseHttpRespBody', () => done()); // Will throw.
       util.handleResp(null, null, null, done);
     });
+
+    it('should surface the response body in the error message when it cannot be JSON-parsed', done => {
+      const unparseableBody = '<html>There was some error</html>';
+
+      util.handleResp(null, null, unparseableBody, err => {
+        if (err === null) {
+          assert.fail('there should be an error');
+        } else {
+          assert.ok(err.message.includes(unparseableBody));
+        }
+
+        done();
+      });
+    });
   });
 
   describe('parseHttpRespMessage', () => {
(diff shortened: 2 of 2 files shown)


test/util.ts
```diff
@@ -377,6 +377,20 @@ describe('common/util', () => {
       stub('parseHttpRespBody', () => done()); // Will throw.
       util.handleResp(null, null, null, done);
     });
+
+    it('should surface the response body in the error message when it cannot be JSON-parsed', done => {
+      const unparseableBody = '<html>There was some error</html>';
+
+      util.handleResp(null, null, unparseableBody, err => {
+        if (err === null) {
```
stephenplusplus: This function body could be:

```js
assert.ifError(err);
assert(err.message.includes(unparseableBody));
done();
```

src/util.ts
```diff
@@ -415,7 +415,9 @@ export class Util {
       try {
         parsedHttpRespBody.body = JSON.parse(body);
       } catch (err) {
-        parsedHttpRespBody.err = new ApiError('Cannot parse JSON response');
+        parsedHttpRespBody.err = new ApiError(
+          `Cannot parse response as JSON: ${body}`
+        );
```
stephenplusplus: Since we're getting fancy, can we put a property on this ApiError instance called `body`?

test/util.ts
Tyler-Murphy: I did it this way to protect against `err` being `null`. Here's what happens otherwise:

```
test/util.ts:386:19 - error TS2531: Object is possibly 'null'.

386         assert.ok(err.message.includes(unparseableBody));
                      ~~~
```

stephenplusplus: Woops, my suggestion didn't make sense at all given what the test is trying to do. Sorry, ignore that!

src/util.ts
Tyler-Murphy: Are you opposed to having it in the message? Or do you want it to be in the `body` property in addition to the message?

I'd like to keep it right in the message. In my experience, only the error message, or the error message and its stack are logged and/or saved by high-level application error handlers for analysis after the fact. Using non-standard error properties makes it more likely that no one will get to see the body.

What do you think?

stephenplusplus: Yes, definitely still in the message. Error messages are often eclipsed, and the user might want to format/display it without our prefix (`Cannot parse response as JSON:`).

Tyler-Murphy: Okay. I see that `ApiError` already has a `response` property. How about adding `handleResp`'s `resp` to the error whenever there's an error, by adding the following to `handleResp`?

```ts
    if (parsedResp.err && resp) {
      parsedResp.err.response = resp
    }
```

If I understand correctly, that would make the body available through `err.response.body`.

stephenplusplus: That seems like a good plan 👍

Tyler-Murphy: Okay, I gave that a try in the latest commits.

PR #450, opened 2019-07-09
fix(ts)!: provide optional typing for set/getMetadata

jkwlui:
Add an additional generic type `U = any` to `ServiceObject` to provide better typings for getMetadata/setMetadata methods.

Currently, inherited `getMetadata` and `setMetadata` return `metadata` object as `any` - users do not get type hints or safety on the metadata object returned by calling these methods.

Upgrading to this release will not cause breaking change because the generic type `U` is defaulted to `any` if not provided by the subclass. Packages that later override the generic type however will cause breaking change as:

1) `metadata` property of the inheriting `ServiceObject` class will be typed to the actual metadata type
2) `getMetadata` and `setMetadata` will invoke callback/ return promise that resolves to the metadata type, which is previously an `any`.
3) `setMetadata` will accept `Partial<U>`, which is the subset of the metadata that the user intends to update, which is previously an `any`.

### DIFF
--- a/src/service-object.ts
+++ b/src/service-object.ts
@@ -366,10 +366,10 @@ class ServiceObject<T = Metadata> extends EventEmitter {
           this.create(...args);
           return;
         }
-        callback!(err, null, metadata as unknown as r.Response);
+        callback!(err, null, (metadata as unknown) as r.Response);
         return;
       }
-      callback!(null, this, metadata as unknown as r.Response);
+      callback!(null, this, (metadata as unknown) as r.Response);
     });
   }
 
--- a/src/service-object.ts
+++ b/src/service-object.ts
@@ -53,10 +53,10 @@ export type GetMetadataOptions = object;
 
 // tslint:disable-next-line:no-any
 export type Metadata = any;
-export type MetadataResponse = [Metadata, r.Response];
-export type MetadataCallback = (
+export type MetadataResponse<T = Metadata> = [T, r.Response];
+export type MetadataCallback<T = Metadata> = (
   err: Error | null,
-  metadata?: Metadata,
+  metadata?: T,
   apiResponse?: r.Response
 ) => void;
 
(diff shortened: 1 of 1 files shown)


src/service-object.ts
```diff
@@ -366,10 +366,10 @@ class ServiceObject extends EventEmitter {
           this.create(...args);
           return;
         }
-        callback!(err, null, metadata as r.Response);
+        callback!(err, null, (metadata as unknown) as r.Response);
```
jkwlui: The type for the third argument is wrong. It's suppose to be a `r.Response` but we're returning the metadata.

JustinBeckwith: I think this was totally wrong.  The last argument shouldn't be of type `r.Response`, it should be `Metadata`.  This was busted to begin with.

jkwlui: In all other methods, the return type is `r.Response`.

callmehiphop: I don't think we've ever actually returned the entire response object though, just the response body?

JustinBeckwith: Exactly.  The original types were a mistake all over.

jkwlui: The third argument should be a `ParsedHttpRespMessage`:

```ts
  /**
   * Sniff an incoming HTTP response message for errors.
   *
   * @param {object} httpRespMessage - An incoming HTTP response message from `request`.
   * @return {object} parsedHttpRespMessage - The parsed response.
   * @param {?error} parsedHttpRespMessage.err - An error detected.
   * @param {object} parsedHttpRespMessage.resp - The original response object.
   */
  parseHttpRespMessage(httpRespMessage: r.Response) {
    const parsedHttpRespMessage = {
      resp: httpRespMessage,
    } as ParsedHttpRespMessage;

    if (httpRespMessage.statusCode < 200 || httpRespMessage.statusCode > 299) {
      // Unknown error. Format according to ApiError standard.
      parsedHttpRespMessage.err = new ApiError({
        errors: new Array<GoogleInnerError>(),
        code: httpRespMessage.statusCode,
        message: httpRespMessage.statusMessage,
        response: httpRespMessage,
      });
    }

    return parsedHttpRespMessage;
  }
```

```ts
export interface ParsedHttpRespMessage {
  resp: r.Response;
  err?: ApiError;
}
```

It's not the response body either.

callmehiphop: Well this is funky. I knew that calling `Service#request` would yield the actual response as a third parameter, I didn't think we returned it anywhere. Taking a closer look it looks like that is not true for `getMetadata`, `setMetadata` and `delete` 😱, however in this case `get` does in fact return the response body as the third parameter.

JustinBeckwith: Heh, gross

PR #455, opened 2019-07-29
feat: add timeout to options and use it in the reqOpts

AVaksman:
add `timeout` to `service` and pass it to reqOpts.
With this `timeout` can be exposed to users in the child dependencies (storage, bigquery...)
Will apply to googleapis/nodejs-storage/issues/733

- [x] Tests and linter pass

### DIFF
--- a/src/service.ts
+++ b/src/service.ts
@@ -69,6 +69,7 @@ export interface ServiceOptions extends GoogleAuthOptions {
   promise?: PromiseConstructor;
   email?: string;
   token?: string;
+  timeout?: number; // http.request.options.timeout
 }
 
 export class Service {
@@ -84,6 +85,7 @@ export class Service {
   authClient: GoogleAuth;
   private getCredentials: {};
   readonly apiEndpoint: string;
+  timeout: number;
 
   /**
    * Service is a base class, meant to be inherited from by a "service," like
@@ -103,6 +105,7 @@ export class Service {
   constructor(config: ServiceConfig, options: ServiceOptions = {}) {
     this.baseUrl = config.baseUrl;
     this.apiEndpoint = config.apiEndpoint;
+    this.timeout = options.timeout!;
     this.globalInterceptors = arrify(options.interceptors_!);
     this.interceptors = [];
     this.packageJson = config.packageJson;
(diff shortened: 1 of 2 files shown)


src/service.ts
```diff
@@ -84,6 +85,7 @@ export class Service {
   authClient: GoogleAuth;
   private getCredentials: {};
   readonly apiEndpoint: string;
+  timeout: number;
```
JustinBeckwith: Wouldn't this be optional?  Or are we setting it to a default value?

src/service.ts
```diff
@@ -69,6 +69,7 @@ export interface ServiceOptions extends GoogleAuthOptions {
   promise?: PromiseConstructor;
   email?: string;
   token?: string;
+  timeout?: number; // http.request.options.timeout
```
JustinBeckwith: @alexander-fenster I'm curious - do we expose a gRPC option for timeout in the GAPIC libraries?

src/service.ts
```diff
@@ -103,6 +105,7 @@ export class Service {
   constructor(config: ServiceConfig, options: ServiceOptions = {}) {
     this.baseUrl = config.baseUrl;
     this.apiEndpoint = config.apiEndpoint;
+    this.timeout = options.timeout!;
```
JustinBeckwith: I'd be nervous using the `!` here because it looks like it can actually be undefined

src/service.ts
alexander-fenster: Not directly, but GAPIC libraries can pass any gRPC option by prefixing it with `grpc.` and it will be passed all the way down to gRPC.

src/service.ts
AVaksman: If `options.timeout` is undefined the [default](https://github.com/googleapis/nodejs-common/blob/master/src/util.ts#L36) value for timeout will be used [later](https://github.com/googleapis/nodejs-common/blob/master/src/util.ts#L717) and not swapped with `undefined` [here](https://github.com/googleapis/nodejs-common/blob/master/src/util.ts#L744)

src/service.ts
AVaksman: I am referencing this [below](https://github.com/googleapis/nodejs-common/pull/455#discussion_r308340988)

PR #480, opened 2019-09-10
feat: return more helpful error when authentication fails

stephenplusplus:
Fixes https://github.com/googleapis/google-auth-library-nodejs/issues/620

Not all API endpoints require authentication, so even when the auth library can't figure out how to generate an API token, we still want to send the API request without one. It might be totally fine.

Or, it might not be, and the API actually requires a token. Fine, then. It returns a 401 with *that API's version of a helpful error.* But, turns out, the auth library has a pretty good version of a helpful error that we'd rather use.

### DIFF
--- a/src/util.ts
+++ b/src/util.ts
@@ -600,6 +600,7 @@ export class Util {
         err: Error | null,
         authenticatedReqOpts?: DecorateRequestOptions
       ) => {
+        const authLibraryError = err;
         const autoAuthFailed =
           err &&
           err.message.indexOf('Could not load the default credentials') > -1;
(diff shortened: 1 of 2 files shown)


src/util.ts
```diff
@@ -651,7 +652,18 @@ export class Util {
           activeRequest_ = util.makeRequest(
             authenticatedReqOpts!,
             reqConfig,
-            callback!
+            // tslint:disable-next-line:only-arrow-functions
```
JustinBeckwith: You can use an arrow function and rest params to do this!
```js
(apiResposeError, ...params) => {
  // ....
  callback!(apiResponseError, ...params);
```

stephenplusplus: Forehead slap.

PR #520, opened 2020-01-23
feat: allow poll interval to be configured

bcoe:
Allow the poll interval for long running operations to be configured.

### DIFF
--- a/src/service-object.ts
+++ b/src/service-object.ts
@@ -97,7 +97,7 @@ export interface ServiceObjectConfig {
    * For long running operations, how often should the client poll
    * for completion.
    */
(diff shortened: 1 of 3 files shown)


src/service-object.ts
```diff
@@ -92,6 +92,12 @@ export interface ServiceObjectConfig {
    * object is Bucket.
    */
   parent: ServiceObjectParent;
+
+  /**
+   * For long running operations, how often should the client poll
+   * for completion.
+   */
+  pollInterval?: number;
```
stephenplusplus: Only idea would be to change to pollIntervalMs to be clear of the expected value.

PR #554, opened 2020-03-25
chore!: drop support for node.js 8

JustinBeckwith:


bcoe:
The Windows CI/CD issue is most likely due to linting being run as a post test step.

We might want to dig into why linting has issues on Windows ... but for now, it might be okay to just drop that post test check.

PR #587, opened 2020-06-17
fix: allow non-JSON body

jkwlui:
I removed the code that throws an `ApiError` if the response body turns out to be non-JSON, a bit weary of the unintended consequences this might have to our downstream libraries. This technically shouldn't occur if the API behaves correctly, i.e. always a JSON body if code is 200-299.

### DIFF
--- a/src/util.ts
+++ b/src/util.ts
@@ -419,9 +419,7 @@ export class Util {
       try {
         parsedHttpRespBody.body = JSON.parse(body);
       } catch (err) {
-        parsedHttpRespBody.err = new ApiError(
-          `Cannot parse response as JSON: ${body}`
-        );
+        parsedHttpRespBody.body = body;
       }
     }
 
--- a/test/util.ts
+++ b/test/util.ts
@@ -392,38 +392,33 @@ describe('common/util', () => {
       util.handleResp(null, null, null, done);
     });
 
-    it('should surface the response body in the error message when it cannot be JSON-parsed', done => {
-      const unparseableBody = '<html>There was some error</html>';
+    it('should handle non-JSON body', done => {
+      const unparseableBody = '<html>Unparseable body.</html>';
 
-      util.handleResp(null, null, unparseableBody, err => {
-        if (err === null) {
-          assert.fail('there should be an error');
-        } else {
-          assert.ok(err.message.includes(unparseableBody));
-        }
-
+      util.handleResp(null, null, unparseableBody, (err, body) => {
+        assert(body.includes(unparseableBody));
         done();
       });
     });
 
-    it('should surface the response body as an error property when it cannot be JSON-parsed', done => {
-      const unparseableBody = '<html>There was some error</html>';
+    it('should include the status code when the error body cannot be JSON-parsed', done => {
+      const unparseableBody = 'Bad gateway';
+      const statusCode = 502;
 
       util.handleResp(
         null,
-        {body: unparseableBody} as r.Response,
+        {body: unparseableBody, statusCode} as r.Response,
         unparseableBody,
         err => {
-          if (err === null) {
-            assert.fail('there should be an error');
-          } else {
-            const response = (err! as ApiError).response;
+          assert(err, 'there should be an error');
+          const apiError = err! as ApiError;
+          assert.strictEqual(apiError.code, statusCode);
 
-            if (!response) {
-              assert.fail('there should be a response property on the error');
-            } else {
-              assert.strictEqual(response.body, unparseableBody);
-            }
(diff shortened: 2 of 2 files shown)


src/util.ts
```diff
@@ -419,9 +419,7 @@ export class Util {
       try {
         parsedHttpRespBody.body = JSON.parse(body);
       } catch (err) {
-        parsedHttpRespBody.err = new ApiError(
-          `Cannot parse response as JSON: ${body}`
-        );
+        parsedHttpRespBody.body = body;
```
JustinBeckwith, line 422: @bcoe this is one of those "technically a breaking change" issues where we're changing the type of the error being returned.  I uh, don't know what the right answer is here.

bcoe: @jkwlui could we potentially look at response headers, and continue to throw if the response is `application/json`?

In general, I wonder if we could be using. response headers rather than waiting for the throw.

jkwlui:
I'm getting more confused on how error handling is done here. I tested with:

```
import {util} from '../src/util';
import * as nock from 'nock';
import * as assert from 'assert';

nock('http://test.com')
    .get('/')
    .reply(400, {error:'not found'});

util.makeRequest({uri: 'http://test.com'}, {}, (err: any, body: any, resp: any) => {
    // expect err.code to be 400, but got undefined
    assert.strictEqual(err.code, 400);
});
```

stephenplusplus:
@jkwlui I added a new test, and that worked:

```js
  describe.only('test', () => {
    nock('http://test.com')
      .get('/')
      .reply(400, {error: {code: 400, errors: [{message: 'not found'}]}});

    it('test', (done) => {
      util.makeRequest({uri: 'http://test.com'}, {}, (err: any, body: any, resp: any) => {
        // expect err.code to be 400, but got undefined
        assert.strictEqual(err.code, 400);
        done();
      });
    })
  });
```

src/util.ts
```diff
@@ -419,9 +419,7 @@ export class Util {
       try {
         parsedHttpRespBody.body = JSON.parse(body);
       } catch (err) {
-        parsedHttpRespBody.err = new ApiError(
```
stephenplusplus, line 422: What is the reason behind ditching this? It could very well be a problem, but I'd like to understand more, so we don't make the same mistake again.

jkwlui: We discovered the API could return plain-text responses in certain 5xx errors.

stephenplusplus: If it's only 500s, why not let it still be an "ApiError"?

jkwlui: I think the problem is the error code is not propagated through when it's plain text.
```javascript
  describe.only('test', () => {
    nock('http://test.com')
      .get('/')
      .reply(503, "<html>Service unavailable.</html>");

    it('test', (done) => {
      util.makeRequest({uri: 'http://test.com'}, {}, (err: any, body: any, resp: any) => {
        // expect err.code to be 400, but got undefined
        assert.strictEqual(err.code, 503);
        done();
      });
    })
  });
```

stephenplusplus: I see now. I think you're right, we can leave out the "Cannot parse as JSON" part. The only change I'd make is to not do anything in the "catch" handler, since `body` is already assigned at the top of the function.

jkwlui: @bcoe also suggested up top that we could perhaps look at the content-type header from the response and keep throwing in the case where content-type is `application/json` and the response isn't parseable. WDYT?

stephenplusplus: I think the only thing this block did was attempt to JSON.parse the body, then return the error body with the prefix "Could not parse..." if it failed. But, it already determines there's an error and creates an ApiError in the previous call to the "parseHttpRespMessage" function. So that original ApiError ends up being overwritten by this one, the "Could not parse..." one, which doesn't have as much information on it (the `code` property). So, I think I'm okay deleting it completely, like you have it now, since it wasn't ever really helping, as far as I can tell.

src/util.ts
stephenplusplus: Here's what a diff should look like given an HTTP response from the request library:

```
HTTP RESPONSE = {
  code: 500
  statusMessage: Internal Service Error
  body: <html><title>error</html>
}
```

```diff
const [err, response] = util.makeRequest('http://...')
err instanceof ApiError === true
- err.message === 'Cannot parse response as JSON: <html><title>error</html>'
+ err.message === 'Internal Service Error'
err.body = '<html><title>error</html>'
err.response === {...raw response object...}
+ err.code === 500
```

jkwlui:
@googlebot rescan

PR #589, opened 2020-06-19
feat: expose method to get request interceptors

stephenplusplus:
This will allow us to get all of the user-defined request interceptors, which is required when we need to make an API request outside of the nodejs-common "request()" method.

### DIFF
--- a/src/service.ts
+++ b/src/service.ts
@@ -138,6 +138,18 @@ export class Service {
   }
 
   /**
+   * Return the user's custom request interceptors.
+   */
+  getRequestInterceptors(): Function[] {
+    // Interceptors should be returned in the order they were assigned.
+    return ([] as Interceptor[]).slice
+      .call(this.globalInterceptors)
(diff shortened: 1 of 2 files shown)


bcoe:
@stephenplusplus is there a specific feature request we'e adding this for out of curiosity? wondering what this will allow us to do.

stephenplusplus:
Yes, sorry, thanks for asking. The Storage library (and conceivably others) will need this to allow the user to define their own HTTP request headers when performing a resumable upload.

To break that down a bit, currently, if a user sets a request interceptor, they receive the raw “reqOpts” they we make each API request with. This lets the user modify them as necessary, including setting a custom header, for example. The Storage library makes most requests through Service.prototype.request(), which knows what to do with the user’s defined interceptors. But for a streaming resumable upload, we don’t use the request method, we use gcs-resumable-upload, which has no idea about the interceptors.

This PR is one of a few that will be needed to get this support into the Storage library.

PR #591, opened 2020-06-25
feat: add ServiceObject#getRequestInterceptors()

stephenplusplus:
Ah, the good old days of PR #589. Young, innocent @stephenplusplus thought just one method to expose the request interceptors would be enough. Until later that night, in the wee hours of the morning, he shot up instantly from his bed with a terrible shout: "What about the service objects?!"

He was right. Those lonely service object interceptors would not be accessible from the current code-- at least, not reasonably. The next day, he submitted a PR to add a new method to ServiceObject by the same name as the previous PR's addition.

This is that PR.

### DIFF
--- a/src/service-object.ts
+++ b/src/service-object.ts
@@ -34,6 +34,8 @@ import {
 export type RequestResponse = [Metadata, r.Response];
 
 export interface ServiceObjectParent {
+  interceptors: Interceptor[];
+  getRequestInterceptors(): Function[];
   requestStream(reqOpts: DecorateRequestOptions): r.Request;
   request(
     reqOpts: DecorateRequestOptions,
@@ -437,6 +439,17 @@ class ServiceObject<T = any> extends EventEmitter {
   }
 
   /**
+   * Return the user's custom request interceptors.
+   */
+  getRequestInterceptors(): Function[] {
+    // Interceptors should be returned in the order they were assigned.
+    const localInterceptors = this.interceptors
+      .filter(interceptor => typeof interceptor.request === 'function')
+      .map(interceptor => interceptor.request);
+    return this.parent.getRequestInterceptors().concat(localInterceptors);
+  }
+
+  /**
    * Set the metadata for this object.
    *
    * @param {object} metadata - The metadata to set on this object.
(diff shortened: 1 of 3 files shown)


src/service-object.ts
```diff
@@ -573,6 +590,6 @@ class ServiceObject<T = any> extends EventEmitter {
   }
 }
 
-promisifyAll(ServiceObject);
+promisifyAll(ServiceObject, {exclude: ['getRequestInterceptors']});
```
stephenplusplus, line 593: *adding a test for this*

JustinBeckwith, line 593: Will this have an impact on downstream consumers of the library?  Or example - let's say in Storage, I inherit from `ServiceObject`, and then call `promisifyAll(Bucket)`.  Would I then need to specify the same list of methods to exclude, or does having it here cover me?  Part of this is that I don't remember if `promisifyAll` operates over inherited methods, or just own properties.

src/service-object.ts
```diff
@@ -34,6 +34,8 @@ import {
 export type RequestResponse = [Metadata, r.Response];
 
 export interface ServiceObjectParent {
+  interceptors: Interceptor[];
+  getRequestInterceptors(): Function[];
```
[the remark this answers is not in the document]
stephenplusplus: The function is executed with the options from the request library we use, and expects to receive the same type back. Trying to write [this function](https://github.com/googleapis/nodejs-common/blob/0467b31cd37dd6829af1e32e4344f5c86d93010d/src/service.ts#L143) had the linter very angry, so the plain "Function[]" made it in with a previous PR. If anyone is able to help get me started by rewriting that function signature in a way that pleases the linter, it would be nice to have it done correctly.

src/service-object.ts
```diff
@@ -34,6 +34,8 @@ import {
 export type RequestResponse = [Metadata, r.Response];
 
 export interface ServiceObjectParent {
+  interceptors: Interceptor[];
```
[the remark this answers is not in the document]
stephenplusplus: The `interceptors` property is meant to be user-accessible (e.g. `bucket.interceptors.push(...)`). This `ServiceObjectParent` is defining the type for the `serviceObject.parent` property, so it's just about listing the methods and properties we use from within the ServiceObject.prototype methods. Sorry if I'm misunderstanding!

src/service-object.ts
stephenplusplus: >  Would I then need to specify the same list of methods to exclude

Nope, it's covered here. It's only working immediately on the methods from the class.

PR #663, opened 2021-05-12
fix: correctly locate projectId from auth library

### LINKED ISSUE #936: keyFilename in BQ constructor is not enough to authenticate API calls
When a user instantiates a client with just a `keyFilename`, an error pops up saying that a project ID is required. 

So, for example: 
```
  const bigquery = new BigQuery({
    keyFilename:
      PATH_TO_KEYFILE,
  });
```

returns:
```
Error: Sorry, we cannot connect to Cloud Services without a project
 ID. You may specify one with an environment variable named
 "GOOGLE_CLOUD_PROJECT".
    at Object.replaceProjectIdToken (/Users/sofialeon/gcp/nodejs-bigquery/samples/node_modules/@google-cloud/projectify/build/src/index.js:46:19)
    at Util.decorateRequest (/Users/sofialeon/gcp/nodejs-bigquery/samples/node_modules/@google-cloud/common/build/src/util.js:484:36)
    at onAuthenticated (/Users/sofialeon/gcp/nodejs-bigquery/samples/node_modules/@google-cloud/common/build/src/util.js:335:53)
    at /Users/sofialeon/gcp/nodejs-bigquery/samples/node_modules/@google-cloud/common/build/src/util.js:381:21
```

In other client libraries, this should be a sufficient method for providing authentication. See [this bug](https://github.com/googleapis/google-auth-library-nodejs/issues/1084) for more details on the investigation.

stephenplusplus:
Fixes https://github.com/googleapis/nodejs-bigquery/issues/936

Previously, the user's project ID was checked against a private property from `google-auth-library`: `_cachedProjectId`. This is not always set, including when a `keyFilename` argument is passed. I found a method from the auth library, `getProjectId()`, which will first search for, then return the most correct project ID.

### DIFF
--- a/src/util.ts
+++ b/src/util.ts
@@ -614,8 +614,7 @@ export class Util {
         }
 
         if (!err || autoAuthFailed) {
-          // eslint-disable-next-line @typescript-eslint/no-explicit-any
-          let projectId = await (authClient as any).getProjectId();
+          let projectId = await authClient.getProjectId();
 
           if (config.projectId && config.projectId !== '{{projectId}}') {
             projectId = config.projectId;
--- a/src/util.ts
+++ b/src/util.ts
@@ -598,7 +598,7 @@ export class Util {
       const callback =
         typeof optionsOrCallback === 'function' ? optionsOrCallback : undefined;
 
-      const onAuthenticated = (
+      const onAuthenticated = async (
         err: Error | null,
         authenticatedReqOpts?: DecorateRequestOptions
       ) => {
@@ -615,7 +615,7 @@ export class Util {
 
         if (!err || autoAuthFailed) {
           // eslint-disable-next-line @typescript-eslint/no-explicit-any
-          let projectId = (authClient as any)._cachedProjectId;
+          let projectId = await (authClient as any).getProjectId();
 
           if (config.projectId && config.projectId !== '{{projectId}}') {
             projectId = config.projectId;
--- a/src/util.ts
+++ b/src/util.ts
@@ -585,7 +585,7 @@ export class Util {
       optionsOrCallback?: MakeAuthenticatedRequestOptions | BodyResponseCallback
     ): void | Abortable | Duplexify {
       let stream: Duplexify;
-      let projectId = config.projectId;
+      let projectId: string;
       const reqConfig = extend({}, config);
       let activeRequest_: void | Abortable | null;
 
@@ -618,7 +618,7 @@ export class Util {
           try {
             authenticatedReqOpts = util.decorateRequest(
               authenticatedReqOpts!,
-              projectId!
+              projectId
             );
             err = null;
           } catch (e) {
--- a/src/util.ts
+++ b/src/util.ts
@@ -585,6 +585,7 @@ export class Util {
       optionsOrCallback?: MakeAuthenticatedRequestOptions | BodyResponseCallback
     ): void | Abortable | Duplexify {
       let stream: Duplexify;
+      let projectId = config.projectId;
       const reqConfig = extend({}, config);
       let activeRequest_: void | Abortable | null;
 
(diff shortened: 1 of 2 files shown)


src/util.ts
```diff
@@ -598,7 +598,7 @@ export class Util {
       const callback =
         typeof optionsOrCallback === 'function' ? optionsOrCallback : undefined;
 
-      const onAuthenticated = (
+      const onAuthenticated = async (
```
JustinBeckwith, line 601: It's a little suspicious that this was switched to an async method, but I don't see an `await` anywhere getting added.  Where is this function called from?

src/util.ts
```diff
@@ -615,7 +615,7 @@ export class Util {
 
         if (!err || autoAuthFailed) {
           // eslint-disable-next-line @typescript-eslint/no-explicit-any
-          let projectId = (authClient as any)._cachedProjectId;
```
JustinBeckwith, line 618: this would be a good time to get rid of the `as any` here, since that method should very much be public and exist on the types now

test/util.ts
```diff
@@ -697,7 +697,9 @@ describe('common/util', () => {
   describe('makeAuthenticatedRequestFactory', () => {
     const authClient = {
       getCredentials() {},
-      _cachedProjectId: 'project-id',
+      getProjectId: () => {
```
JustinBeckwith, line 700: Since this is async in the library, should it be an async function here?

src/util.ts
stephenplusplus: This was just to enable me to use an await inside to get the project ID, without using the whole promise/callback thing.

src/util.ts
stephenplusplus: Good call, done!

test/util.ts
stephenplusplus: Makes sense to me! Changed 👍🏻

src/util.ts
JustinBeckwith: Sure, but now when this method is called, it won't execute synchronously.  Where we invoke `onAuthenticated`, I would expect to now require an `await`?

stephenplusplus: Sorry, that was definitely a mistake! I've made an update to consolidate some code and embrace the promise chain :)

sofisl:
@stephenplusplus, excited to get this PR merged! Would you mind taking a look at some of our comments, excited to look over and merge soon.

src/util.ts
```diff
@@ -614,17 +615,10 @@ export class Util {
...
-
-          if (config.projectId && config.projectId !== '{{projectId}}') {
-            projectId = config.projectId;
-          }
-
           try {
             authenticatedReqOpts = util.decorateRequest(
               authenticatedReqOpts!,
-              projectId
+              projectId!
```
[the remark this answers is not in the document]
stephenplusplus: Sorry I missed this comment! I was able to remove the `!` in the last commit. PTAL and let me know if I overlooked anything. Thanks!

src/util.ts
```diff
@@ -618,7 +618,7 @@ export class Util {
           try {
             authenticatedReqOpts = util.decorateRequest(
               authenticatedReqOpts!,
-              projectId!
+              projectId
```
stephenplusplus: This code is inside of the callback function, `onAuthenticated`, executed after `authClient.getProjectId()`. That is when the `projectId` is assigned using what the authClient says is the correct project ID.

src/util.ts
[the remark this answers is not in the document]
stephenplusplus: Woops, I noticed I left an edge case out. Fix incoming!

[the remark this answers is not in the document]
stephenplusplus: Woops, I noticed I left an edge case out. Fix incoming!

tmatsuo:
What's remaining on this PR?

stephenplusplus:
I think we just need a @sofisl review.

bcoe (approved):

PR #674, opened 2021-06-16
feat: began retry configuration implementation

shaffeeullah:
Custom retry strategy behavior. Will merge all retry PRs into the shaffeeullah/storageRetryStrategy branch and then merge that branch into master at the end. This avoids one giant PR.

There is one test currently failing. I believe it is because the options that the `makeRequest` function is pulling from `retry-request` are not the options that we're looking for. This change is pending the `retry-request` library update.

### DIFF
--- a/src/util.ts
+++ b/src/util.ts
@@ -743,8 +743,7 @@ export class Util {
       config.autoRetry !== undefined &&
       config.retryOptions?.autoRetry !== undefined
     ) {
-      throw new ApiError('autoRetry is deprecated. Use retryOptions.autoRetry instead.')
-      );
+      throw new ApiError('autoRetry is deprecated. Use retryOptions.autoRetry instead.');
     } else if (config.autoRetry !== undefined) {
       autoRetryValue = config.autoRetry;
     } else if (config.retryOptions?.autoRetry !== undefined) {
@@ -754,8 +753,7 @@ export class Util {
     const MAX_RETRY_DEFAULT = 3;
     let maxRetryValue = MAX_RETRY_DEFAULT;
     if (config.maxRetries && config.retryOptions?.maxRetries) {
-      throw new ApiError('maxRetries is deprecated. Use retryOptions.maxRetries instead.')
-      );
+      throw new ApiError('maxRetries is deprecated. Use retryOptions.maxRetries instead.');
     } else if (config.maxRetries) {
       maxRetryValue = config.maxRetries;
     } else if (config.retryOptions?.maxRetries) {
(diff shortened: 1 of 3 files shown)


bcoe:
It's exciting to see this work coming together 😄

stephenplusplus:
@shaffeeullah I have opened the retry-request PR here: https://github.com/stephenplusplus/retry-request/pull/35

src/util.ts
```diff
@@ -727,9 +737,36 @@ export class Util {
     config: MakeRequestConfig,
     callback: BodyResponseCallback
   ): void | Abortable {
+    const AUTO_RETRY_DEFAULT = true;
```
ddelgrosso1, line 740: Should these constants also reside at the class level with some comments / docs?

shaffeeullah: Good callout!

ddelgrosso1 (approved):
LGTM

src/util.ts
```diff
@@ -727,13 +753,41 @@ export class Util {
     config: MakeRequestConfig,
     callback: BodyResponseCallback
   ): void | Abortable {
+    let autoRetryValue = AUTO_RETRY_DEFAULT;
+    if (
+      config.autoRetry !== undefined &&
+      config.retryOptions?.autoRetry !== undefined
```
stephenplusplus, line 759: Will `config.retryOptions` always be an object (never undefined)?

shaffeeullah: It will not always be an object. In this test, it is undefined. https://github.com/googleapis/nodejs-common/blob/9bd6d9e16ff0102b526860ecbc2f33985bca4959/test/util.ts#L1420

stephenplusplus: Nice, thanks!

src/util.ts
```diff
@@ -727,13 +753,41 @@ export class Util {
...
+    ) {
+      throw new ApiError(
```
bcoe, line 761: If I'm understanding this logic, we through if a customer sets both `autoRetry` and `retryOptions.autoRetry`?

Is there any risk that when this is deployed existing customer applications are somehow setting both these parameters?

bcoe (approved):
LGTM, only question is, is there any chance that existing customer applications would hit one of the code paths that throws?

For instance, did we direct customers to set `config.autoRetry` manually at any point, and storage will now be setting `config.retryOptions.autoRetry` automatically?

src/util.ts
shaffeeullah: All applications using nodejs-common should not be using these fields because they did not exist prior to this PR. Storage handles this case directly https://github.com/googleapis/nodejs-storage/pull/1474/files

PR #675, opened 2021-06-22
feat: implemented retry function

shaffeeullah:
Allows users to pass in a custom function that returns true if the error should be retried and false otherwise

### DIFF
--- a/src/util.ts
+++ b/src/util.ts
@@ -324,7 +324,6 @@ export interface MakeRequestConfig {
   shouldRetryFn?: (response?: r.Response) => boolean;
 
   retryableErrorFn?: (err: ApiError) => boolean; //TODO: refactor into RetryConfig
-
 }
 
 export class Util {
@@ -731,8 +730,10 @@ export class Util {
     callback: BodyResponseCallback
   ): void | Abortable {
     const DEFAULT_RETRYABLE_ERR_FN = util.shouldRetryRequest;
-    const RETRYABLE_ERR_FN = config.retryOptions.retryableErrorFn ? config.retryOptions.retryableErrorFn : DEFAULT_RETRYABLE_ERR_FN;
-    
+    const RETRYABLE_ERR_FN = config.retryOptions.retryableErrorFn
+      ? config.retryOptions.retryableErrorFn
+      : DEFAULT_RETRYABLE_ERR_FN;
+
     const options = {
       request: teenyRequest.defaults(requestDefaults),
       retries: config.autoRetry !== false ? config.maxRetries || 3 : 0,
--- a/src/util.ts
+++ b/src/util.ts
@@ -538,6 +538,7 @@ export class Util {
    * @return {boolean} True if the API request should be retried, false otherwise.
    */
   shouldRetryRequest(err?: ApiError) {
+    console.log('hereshouldretryrequst')
     if (err) {
       if ([408, 429, 500, 502, 503].indexOf(err.code!) !== -1) {
         return true;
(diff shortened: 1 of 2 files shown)


test/util.ts
```diff
@@ -1214,6 +1214,36 @@ describe('common/util', () => {
         config.shouldRetryFn!();
       };
     }
+    const errorMessage = 'Error.';
+    const customRetryRequestFunctionConfig = {
+      retryOptions: {
+        retryableErrorFn: function (err: ApiError) {
+          if (err.message === errorMessage) {
```
ddelgrosso1, line 1221: Perhaps nitpick-y but this could be made more succinct by just saying `return err.message === errorMessage`.

shaffeeullah: so smart

ddelgrosso1 (approved):
LGTM

src/util.ts
```diff
@@ -325,6 +325,7 @@ export interface RetryOptions {
   maxRetryDelay?: number;
   autoRetry?: boolean;
   maxRetries?: number;
+  retryableErrorFn?: (err: ApiError) => boolean;
```
stephenplusplus, line 328: Would it make sense to just call this `shouldRetryRequest`? Then it would be consistent with the naming and functionality of retry-request and the other transient options.

shaffeeullah: There's already this function called `shouldRetryRequest` https://github.com/googleapis/nodejs-common/blob/04d43b63aabaebb6bdfcc4b2b4f760703249edfb/src/util.ts#L513 and this parameter called `shouldRetryFn` (which is a different function with a different signature than this function) https://github.com/googleapis/nodejs-common/blob/04d43b63aabaebb6bdfcc4b2b4f760703249edfb/src/util.ts#L324

I thought the naming of `shouldRetryRequest` would be confusing in the context of this repo. I chose `retryableErrorFn` because it calls out `Error` which is the main purpose. Open to other opinions here though. @bcoe & @ddelgrosso1 what are your thoughts?

stephenplusplus: Sorry, I meant to suggest renaming `retryableErrorFn` to `shouldRetryFn`.

```ts
    shouldRetryFn(httpRespMessage: r.Response) {
        const err = util.parseHttpRespMessage(httpRespMessage).err;
        if (config.retryOptions?.retryableErrorFn) {
          return err && config.retryOptions?.retryableErrorFn(err);
        }
        return err && util.shouldRetryRequest(err);
      }
```

I can't find where `MakeRequestConfig.shouldRetryFn` gets used? I believe that should be the option that, if defined, overrides `util.shouldRetryRequest`.

shaffeeullah: `MakeRequestConfig.shouldRetryFn` and `retryableErrorFn` have different signatures. `retryableErrorFn` takes in an error directly. This removes duplicate code (the rest of the function would be the same) and also makes it easier for other APIs (like Storage) to use this functionality without depending on `retry-request` for `r.Response`

bcoe (approved):

src/util.ts
shaffeeullah: @stephenplusplus What do you think about this design?

stephenplusplus: I think I'm stuck on where `MakeRequestConfig.shouldRetryFn` gets used. I see we expose the option, but it doesn't seem like we're doing anything with it. So then having `MakeRequestConfig.retryableErrorFn` would add to my confusion, if I was looking at my options... would I want to use "shouldRetryFn" or "retryableErrorFn" to confirm a retry? Do I define both, to handle two scenarios? One for when I get a raw response and one when I get an error?

I could be wrong, but it seems like the intention of retry-request's "shouldRetryFn" function is being manipulated a little around our interface and previously defined types. From retry-request's perspective, `shouldRetryFn` is all that's provided so that the user can take the response and determine if retry-request should retry the request. I think it's okay for us to sanitize the response into our custom Error format, and still consider the feature the same.

But since I'm just trying to catch up, I fully trust if you see the changes in this PR as the best way to offer the best behavior and the most simplicity. I could definitely be forgetting something, and apologies if I'm making things needlessly complicated. Feel free to carry on if that's the case!

shaffeeullah: I agree with you that this is confusing. The `MakeRequestConfig.shouldRetryFn` option exists on the interface, but is never actually read. I think we should look into deprecating this option (cc @bcoe ) at some point or better documenting it. From the perspective of storage and other libraries that could potentially use this in the future, I think breaking it down to error input, boolean output is much simpler.

PR #704, opened 2021-11-01
feat: Add retry support for 504 (Gateway Timeout)

### LINKED ISSUE #1635: Add retry logic to file.download because it fails with 504 often
**Is your feature request related to a problem? Please describe.**

My `file.download` calls often fail with 504 gatewayTimeout.
Guesstimating that this happens 1-5% of the time, over 1000s of calls per day.

 **Describe the solution you'd like**

It would be nice if `file.download` had the same kind of retry handling that `file.save` has.

 **Describe alternatives you've considered**

I have implemented my own retry logic on top of this library, and it's similar to how the retry logic is implemented for `file.save`.

---

Here's the relevant part of the error object I'm seeing:

```javascript
 ApiError {
  code: 504,
  errors: [ { message: '', domain: 'global', reason: 'gatewayTimeout' } ],
  response: PassThrough {
    _readableState: ReadableState {
      objectMode: false,
...
```

danielbankhead:
Fixes a downstream issue https://github.com/googleapis/nodejs-storage/issues/1635 by adding support for gateway timeouts - which is also compliant with https://cloud.google.com/storage/docs/retry-strategy#idempotency-operations

Thank you for opening a Pull Request! Before submitting your PR, there are a few things you can do to make sure it goes smoothly:
- [ ] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/nodejs-common/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea
- [ ] Ensure the tests and linter pass
- [ ] Appropriate docs were updated (if necessary)

Fixes https://github.com/googleapis/nodejs-storage/issues/1635 🦕

### DIFF
--- a/src/util.ts
+++ b/src/util.ts
@@ -544,7 +544,7 @@ export class Util {
    */
   shouldRetryRequest(err?: ApiError) {
     if (err) {
-      if ([408, 429, 500, 502, 503].indexOf(err.code!) !== -1) {
+      if ([408, 429, 500, 502, 503, 504].indexOf(err.code!) !== -1) {
         return true;
       }
 
--- a/test/util.ts
+++ b/test/util.ts
@@ -1204,6 +1204,12 @@ describe('common/util', () => {
       assert.strictEqual(util.shouldRetryRequest(error), true);
     });
 
+    it('should return true with error code 504', () => {
+      const error = new ApiError('504');
+      error.code = 504;
+      assert.strictEqual(util.shouldRetryRequest(error), true);
+    });
+
     it('should detect rateLimitExceeded reason', () => {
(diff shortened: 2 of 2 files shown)


danielbankhead:
Added `labels/do not merge` as this may be a breaking change for downstream customers (need to investigate)

JustinBeckwith (approved):
From a breaking change perspective, I think this is fine and won't harm anything.  From a practical standpoint though, we need to make 100% sure GCS isn't the cause of these responses.  It's very likely a users upstream proxy causing the problem, and there's only so much dancing around that we can really do.

PR #707, opened 2021-11-03
fix: disable retrying no response if autoRetry is set to false

ddelgrosso1:
Thank you for opening a Pull Request! Before submitting your PR, there are a few things you can do to make sure it goes smoothly:
- [ ] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/nodejs-common/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea
- [ ] Ensure the tests and linter pass
- [ ] Appropriate docs were updated (if necessary)

Fixes #<issue_number_goes_here> 🦕


src/util.ts
```diff
@@ -784,6 +784,7 @@ export class Util {
     const options = {
       request: teenyRequest.defaults(requestDefaults),
       retries: autoRetryValue !== false ? maxRetryValue : 0,
+      noResponseRetries: autoRetryValue !== false ? maxRetryValue : 0,
```
shaffeeullah, line 787: ah wow this is a separate config in the retry request library?

shaffeeullah (approved):

src/util.ts
ddelgrosso1: Yes, `retry-request` has both https://github.com/stephenplusplus/retry-request/blob/f06f6351bd1f4d635804fb1c77e14984f2c9b2e4/index.js#L33 and https://github.com/stephenplusplus/retry-request/blob/f06f6351bd1f4d635804fb1c77e14984f2c9b2e4/index.js#L9

JustinBeckwith (approved):

PR #732, opened 2022-02-17
feat: Support `AuthClient` for `authClient`

danielbankhead:


bcoe (approved):
This is looking great.

PR #753, opened 2022-06-10
fix: Requests Respect `config.projectIdRequired` === `false`

### LINKED ISSUE #711: Regular GCS uploads are failing with "Unable to detect a Project Id in the current environment."
# Basic info
* Node version: 14.16.1
* @google-cloud/storage version: 5.16.1
* @google-cloud/common version: 3.8.1 (as @google-cloud/storage dependency)

# Issue
When using GCS regular uploads with a client_email and private_key it fails with the following error:
```
Unable to detect a Project Id in the current environment.
To learn more about authentication and Google APIs, visit: 
https://cloud.google.com/docs/authentication/getting-started
```
Though there is no need of setting up projectId explicetly - even Storage class set projectIdRequired to false (https://github.com/googleapis/nodejs-storage/blob/c36525402da8e748971473b1cdd2423e8fd953e1/src/storage.ts#L628)

When upload is resumable (with same credentials) projectId is not required, same should apply for regular upload as well.

Code snippet of upload:
```
            let gcs = require('@google-cloud/storage');
            let storage = new gcs.Storage({credentials: {client_email, private_key}});
            src_stream.pipe(
                storage.bucket(bucket).file(filename)
                .createWriteStream({resumable: false, contentType: content_type}),
```

# Proposal
Instead of:
https://github.com/googleapis/nodejs-common/blob/a7e4ee228291e8ec46f4077ee2d73124c106c94c/src/util.ts#L701
skip resolving projectId if projectIdRequired is false:
```
                !config.projectIdRequired || config.projectId && config.projectId !== '{{projectId}}'
```

danielbankhead:
Fixes #711  🦕

**Important**: Will need to port fix to https://github.com/googleapis/nodejs-storage/

### DIFF
--- a/src/util.ts
+++ b/src/util.ts
@@ -30,6 +30,7 @@ import {Duplex, DuplexOptions, Readable, Transform, Writable} from 'stream';
 import {teenyRequest} from 'teeny-request';
 
 import {Interceptor} from './service-object';
+import {DEFAULT_PROJECT_ID_TOKEN} from './service';
 
 // eslint-disable-next-line @typescript-eslint/no-var-requires
 const duplexify: DuplexifyConstructor = require('duplexify');
@@ -596,7 +597,7 @@ export class Util {
     config: MakeAuthenticatedRequestFactoryConfig
   ) {
     const googleAutoAuthConfig = extend({}, config);
-    if (googleAutoAuthConfig.projectId === '{{projectId}}') {
+    if (googleAutoAuthConfig.projectId === DEFAULT_PROJECT_ID_TOKEN) {
       delete googleAutoAuthConfig.projectId;
     }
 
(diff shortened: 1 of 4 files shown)


src/util.ts
```diff
@@ -711,23 +728,55 @@ export class Util {
...
-            // authentication. (ex: connecting to a local Datastore server)
-            new Promise(resolve => resolve(reqOpts))
-          : authClient.authorizeRequest(reqOpts),
-      ])
-        .then(([_projectId, authorizedReqOpts]) => {
-          projectId = _projectId as string;
-          onAuthenticated(null, authorizedReqOpts as DecorateRequestOptions);
-        })
-        .catch(onAuthenticated);
+      const prepareRequest = async () => {
```
danielbankhead, line 731: Expanded this block for readability.

ddelgrosso1:
Overall I think the code changes look good. However, what is the plan for testing this as to not potentially impact consumers in case this causes unforeseen issues?

danielbankhead:
@ddelgrosso1 I think this is overall safe - here's an overview:
- if `projectIdRequired !== false`, the codepaths are the same as before
- if `projectIdRequired === false` it will attempt to use the default project id placeholder (which throws if a project id is required), and will fallback to checking for a project id (preserving existing behavior)

Additionally, all API calls are in the same order as before

bcoe (commented):
Awesome thanks for upstreaming this work @danielbankhead 👏 

Will defer to @ddelgrosso1 for review.

ddelgrosso1 (approved):

PR #813, opened 2024-04-15
fix: replace `ent` with `html-entities`

### LINKED ISSUE #812: Package `ent` isn't maintained anymore, and using deprecated `punycode`
The `ent` package is still using the native `punycode` module which used to be bundled in Node.js < 21.

Starting in Node.js 21, this native module was deprecated (see [official Node.js docs](https://nodejs.org/api/punycode.html)), and we now have to use the [userland `punycode`](https://www.npmjs.com/package/punycode).

And the following warning shows:
```
(node:1) [DEP0040] DeprecationWarning: The `punycode` module is deprecated. Please use a userland alternative instead. 
```

Since the latest version of `ent` is 9-years-old and [the repository was removed from GitHub](https://github.com/substack/node-ent), I think we could replace `ent` with another heavily-used package, e.g. [`html-entities`](https://www.npmjs.com/package/html-entities).

#### Environment details

  - OS: `macOS`
  - Node.js version: `v21.7.3`
  - npm version: `10.5.0`
  - `@google-cloud/common` version: `5.0.1`

#### Steps to reproduce

  1. `npm install @google-cloud/common`
  2. `node --trace-deprecation -e "require('@google-cloud/common')"` (on Node.js 21)

I'll make a PR.

Drarig29:
Fixes #812 

This PR replaces `ent` with `html-entities`.

BEGIN_COMMIT_OVERRIDE
fix: replace ent with html-entities
Release-As: 5.0.2
END_COMMIT_OVERRIDE

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,15 +36,14 @@
     "@google-cloud/promisify": "^4.0.0",
     "arrify": "^2.0.1",
     "duplexify": "^4.1.1",
-    "ent": "^2.2.0",
     "extend": "^3.0.2",
     "google-auth-library": "^9.0.0",
+    "html-entities": "^2.5.2",
     "retry-request": "^7.0.0",
     "teeny-request": "^9.0.0"
   },
   "devDependencies": {
     "@compodoc/compodoc": "^1.1.11",
-    "@types/ent": "^2.2.1",
     "@types/extend": "^3.0.1",
     "@types/mocha": "^9.0.0",
     "@types/mv": "^2.1.0",
--- a/src/util.ts
+++ b/src/util.ts
@@ -20,7 +20,7 @@ import {
   replaceProjectIdToken,
   MissingProjectIdError,
 } from '@google-cloud/projectify';
-import * as ent from 'ent';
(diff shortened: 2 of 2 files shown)


sofisl:
Unfortunately we can't accommodate this request because Node 18 is required for this package, and we're on Node 14. If there's another dependency that would do the trick that supports Node 14, that would work. Or, we can wait until our work to upgrade to Node 16, which should be in the next few months.

Drarig29:
@sofisl `html-entities` has zero dependencies.

The issues actually comes from one of the `devDependencies` in this repository:
https://github.com/googleapis/nodejs-common/blob/926235906e59324e100c07c7aae4175a7d0255e0/package.json#L46

I have the same error on the `main` branch if I run the same command as in the CI **on Node 16**:
```bash
npm install --production --engine-strict --ignore-scripts --no-package-lock
```

So the issue doesn't come from the PR

Drarig29:
@sofisl here is a PR to fix the bug: https://github.com/googleapis/nodejs-common/pull/814

sofisl (approved):

PR #834, opened 2025-02-28
chore!: Upgrade to Node version 18

danieljbruce:


migrate-to-node-18.cjs
```diff
@@ -0,0 +1,90 @@
+const path = require('path');
```
sofisl, line 1: you'll want to delete this file

sofisl (approved):
LGTM after extra file is deleted.

